PR #3, opened 2022-09-24
With Terrain rewrite both flags must be set, of non - will default to…

DjWarmonger:
For https://github.com/vcmi/vcmi/pull/963


Nordsoft91:
@DjWarmonger , please increment mod version too

PR #21, opened 2023-04-17
Feature/Refactor HotA Artifacts mod

krs0:
First pass done. Relocated files, rearranged json contents.


IvanSavenko:
- Personal opinion: don't think that such changes are strictly speaking needed, but no objections against them.
- Nice that interference bonuses got cleanup. Assuming that they do work? You might also want to look into skill itself and maybe - hero specialties
- Images can be batch-converted into .png's to slightly reduce size. No need to bother with manual conversion though.

Nordsoft91:
is it for 1.3?

krs0:
Yes 1.3.. mainly moving things around...updating jsons to latest bonuses but also 1-2 bugs ...
Cannot really test too well due to buggy artifacts 😔
But as agreed with Ivan will integrate today what I have up to now in 1.3

krs0:
I am finished for the moment. Can be merged. Prob in 1.3 branch?

hota/Mods/Artifacts/Content/config/hota/artifacts/goldenGoose.json
```diff
@@ -1,23 +1,31 @@
...
+		],
 		"bonuses" : [
 			{
-				"subtype" : "resource.gold",
 				"type" : "GENERATE_RESOURCE",
-				"val" : 4750,
-				"valueType" : "BASE_NUMBER"
+				"subtype" : "resource.gold",
+				"valueType" : "BASE_NUMBER",
+				"val" : 7000
```
IvanSavenko: Recheck I guess but old version should be correct:
- In addition to component effects brings 4750 Gold gold/day, for a total bonus of 7000 Gold gold/day.
- Component effects: bring 2250 Gold gold/day.

So, artifact *itself* give 4750 gold. But due to component effects, total income is 7000

hota/Mods/Artifacts/Content/config/hota/artifacts/capeOfSilence.json
```diff
@@ -1,22 +1,26 @@
...
+		"class" : "MAJOR",
+		"value" : 7500,
 		"bonuses" : [
-			{ 
-				"propagator": "BATTLE_WIDE",
-				"propagationUpdater" : "BONUS_OWNER_UPDATER",
+			{
 				"type" : "LEVEL_SPELL_IMMUNITY", 
-				"val" : 1 
+				"val" : 2,
```
IvanSavenko, line 10: Not sure how this bonus works, but it does block *both* level 1 and level 2 spells now. Correct?

hota/Mods/Artifacts/Content/config/hota/artifacts/goldenGoose.json
krs0: Maybe I do not understand something...
This is not 'new artifact' that replaces completely components and just blocks some slots? It performs additive calculations for components?

hota/Mods/Artifacts/Content/config/hota/artifacts/capeOfSilence.json
krs0: I looked for red dragons as a model and there it was:
```
			"spellImmunity" :
			{
				"type" : "LEVEL_SPELL_IMMUNITY",
				"val" : 3
			}
```
should work the same. will recheck all once artifacts work again.

hota/Mods/Artifacts/Content/config/hota/artifacts/goldenGoose.json
IvanSavenko: Yep. Bonuses of individual components are still active.
Normally component and composite effects are different.
For example Angelic Alliance only defines alignment mixing and prayer - bonuses to primary skills still come from components:
https://github.com/vcmi/vcmi/blob/develop/config/artifacts.json#L1919
Golden Goose is a bit special here since it has same effect (gold production) as its components

krs0: Ok will fix that.
Sounds strange this approach/limitation... to calculate hidden in the background, individual bonuses from components. If provided as a shortcut/option... sure, but not default.
I also wonder about readability/extensibility of the code implementing this...

In this case, I as a modder do not care about those arithmetics. I care that golden goose gives 7000g. Which is simple and straight forward.

Not to mention that this should be a fused artifact, in the end. No  more components no nothing, just the goose :)

IvanSavenko: You do have a point, but this is how all H3 composite artifacts work - they only extend effect.
H3 tooltips also only state effect from composite artifact. Effects from individual component are omitted from description.

It is indeed less confusing than Golden Goose - since H3 composites provide effects different from their components, so no need for arithmetic.

krs0: ok reverted and proposed one sentence for clarification. If ok than this could be merge from my side.

PR #22, opened 2023-05-14
Reorganize submods

IvanSavenko:
Massive PR as far as git is concerned, please look directly in my branch if you wish to review something:
https://github.com/IvanSavenko/horn-of-the-abyss/tree/vcmi-1.3/hota/Mods

All non-json files were re-imported, using current hota release as base.
Importing is done using shell script to reduce manual work if we need to update mod, e.g. after hota release. Will upload said script a bit later, separately.

List of changes as part of re-import:
- all bmp's are now imported as png to reduce file size
- files that use new 'd32' def format from hota are imported as json + list of png's
- files that no longer exist in hota but present in our mod were removed
- some files that might not be usable in vcmi are skipped and not included (mostly UI changes)

Note that we still use *some* custom content that is not present in hota. For example, commander for wog and some bitmaps were manually converted to format suitable for vcmi

There should be no/minimal changes in functionality due to reimport. Most notable differences are:
- VCMI now uses new puzzle map for Cove, apparently was changed at some point
- Hero portraits for Cove have also been adjusted at some point
- Some decorative obstacles for map have been changed to match hota

Second part involved reorganization of hota submods into more logical scheme and one that would be more suitable for hota h3m support - for example, confux new building (phoenix horde) is now a separate submod to allow use of hota h3m without hota balance submod. Another change is that all submods now use camelCase naming scheme instead of alternating between "xxx-yyy" or "xxx yyy" or "xxxYyy" schemes.

List of changes in submods (other than rename to camelCase):
- cannon: contains cannon, extracted from Cove
- cove: cannon, interference and neutral creatures have been moved to other submods
- gameBalance: renamed from 'HotA balance', all graphical fixes have been moved to graphicalFixes, conflux horde moved to separate mod
- confluxVaultofAshes: contains Phoenix horde for Conflux. Extracted from balance mod since horde is needed for h3m support, unlike rest of balance changes
- interference: contains interference skill itself and former 'interferenceHeroes' submod
- mapObjects: merged version of 'HotA banks' of 'HotA configurable objects'
- neutralCreatures: contains 4 new neutrals, extracted from Cove
- graphicalFixes: renamed from SoD new defs
- mapDecorations: renamed from 'HotA decorations'
- terrains: renamed from 'New Terrains'


misiokles:
It is possible to implement shooting visuals effects directly into creature jsons? Now it's a workaround by using dump spell_after_attack. Which is fine, but can be confusing, because names of these spells are visible in creature ability window.

IvanSavenko:
@misiokles Will think about it, but not as part of this pull request, it already too large.

PR #27, opened 2023-07-02
Json tweaks for newCreatureAnimations, cove heroes and creatures and terrains

### LINKED ISSUE #26: Reorganize and fix HotA mod files
General purpose of this issue is to update and improve the formatting of existing HotA mod. And some minor fixes along the way.

- [x] Reorganize new heroes files (Beatrice, Ranloo, Kinkeria, Giselle)

- [x] Move gendered hero sprites into **newGraphics** mod

- [x] Update Cove moat to use vcmi-1.3 former (moat spell)

- [x] Look into deprecated building ID's warnings in Cove

- [x] Anabel specialty (Pirates) should use short form in 1.3, since bug with multiple upgrades is now fixed

- [x] (?) Move terrains into individual submods to allow their separate usage

krs0:
The purpose of this PR is to give a little more structure and consistency to HotA cove mod.
Fixes: https://github.com/vcmi-mods/horn-of-the-abyss/issues/26


IvanSavenko:
Considering that our modding system is conflict-prone as far as filesystem goes, please use following layout for consistency:
- for all config files from a submod:
`[submodName]/config/hota[SubmodName]/*`
- for all new graphical files:
`[submodName]/data/hota/*`
`[submodName]/sprites/hota/*`
- for all graphical files that should replace H3 files:
`[submodName]/data/*`
`[submodName]/sprites/*`

Meaning, all files that come with mod should have "hota" in their path

Also, personal opinion / not going to insist but I'd rather avoid too deep directory structure unless there is actually large amount of files. E.g
`castle/newCreatureAnimations.json`
can be replaced with
`castleCreatureAnimations.json`
This avoids directories with 1 files and makes editing all files one-by-one easier, since you don't have to constantly change directories in file browser.

krs0:
for all config files from a submod:
`[submodName]/config/hota[SubmodName]/*`
Agree

for all new graphical files:
```
[submodName]/data/hota/*
[submodName]/sprites/hota/*
```
Agree

for all graphical files that should replace H3 files:
```
[submodName]/data/*
[submodName]/sprites/*
```
In case the config file is remade, it would be clearer to use a logical structure like:
`Mods\newGraphics\Content\sprites\hota\castle\heroClasses\cleric`
no need to clump everything together.

maybe even use submod name explicitly:
`Mods\newGraphics\Content\sprites\hota\newGraphics\castle\heroClasses\cleric`

IvanSavenko:
@krs0 
Please, don't create too many directories. It might look "nice" but a pain to edit.
Want to check all 5-10 json's in mod? Now open 10-20 directories with 1 file in them.

As I said before:

for all config files from a submod:
`[submodName]/config/hota[SubmodName]/*`
for all new graphical files:
`[submodName]/data/hota/*`
`[submodName]/sprites/hota/*`
for all graphical files that should replace H3 files:
`[submodName]/data/*`
`[submodName]/sprites/*`

IMO subdirectories should be created only when number of files grows too much.

There is no point in having files like
"config/hota/heroes/cove/navigator/anabel.json"
"data/hota/heroes/cove/navigator/icons/anabel/specialty/large.png"

There will be never a file conflict as long as there is "hota" directory.
And file "anabelSpecialtyLarge.png" is way more convenient to find & edit than "icons/anabel/specialty/large.png"

krs0:
I get your point, but I have some remarks:

Right now Cove is a little messy. You find stuff all over the place. I try to reorganize it, give it some shape and in the end if folders are too much, I can mass replace them with long file names, no problem.

Cove subfolder has to be added. When a new faction will be added to HotA ... having a creatures folder will all creatures from both faction will not do.

Files like AH18_.def if not put in the right folder structure will be up to guess if navigator of captain hero class or if combat or adventure map .def

Consistency would be desireble. If all sounds for a creature are found in one folder, I expect all icons for one creature to be in one folder as well, and not to use the longFileName trick.

I do not really see a big problem/difference when editing, given the tools that we have lately.

krs0:
so lets take them 1 by one: I start with the easy one for me: 
for configuration files... is this ok?
image

krs0:
And another no brainer for me:
sounds folder:
image

krs0:
And for Sprites:
Ignore pls what is not under cove folder, that is WIP
image

Here I have added a creatures folder because creature files were mixed up with town buildings, adventure map stuff and what not and were hard to distinguish

IvanSavenko:
@krs0 
Sorry for delay.

> Cove subfolder has to be added. When a new faction will be added to HotA ... having a creatures folder will all creatures from both faction will not do.

Actually not a problem for *creatures*. After all, it is not like factory will get another nymphs or pirates. However true for some other files, e.g. town buildings.

Ideally I'd like to add filesystem isolation for mods, allowing mods to use minimalistic file layout without fear of conflict with mod you know nothing about. But don't have this yet.

> Files like AH18_.def if not put in the right folder structure will be up to guess if navigator of captain hero class or if combat or adventure map .def

Rename it then to proper name? E.g. `navigatorAdventureMap.def`

> Consistency would be desireble. If all sounds for a creature are found in one folder, I expect all icons for one creature to be in one folder as well, and not to use the longFileName trick.

Personally I'd rather put both sounds & icons in common directories, e.g. `sounds/hota/*.wav`. It is not like hota will get two different sets of sounds for pirates

> I do not really see a big problem/difference when editing, given the tools that we have lately.

If this is OK with everyone who edits hota - then OK.

> so lets take them 1 by one: I start with the easy one for me:
> for configuration files... is this ok?

personally I would rather use `config/hotaCove/objectType/*.json`. No need for separate "hota" directory since it will never contain anything other than Cove

> And another no brainer for me:
> sounds folder:

Personally don't see any need for complex directory structure inside `sounds/hota/`. Number of files is rather low and we already kind of force rather deep directory layout (e.g. `hota/Mods/cove/Content/sounds/hota` - 6(!) directories deep just to find a file.
Also keep in mind that hota sometimes reuses sounds for different objects. I think one (but far from only) example is Fangarm dwelling sound that is the same file as sound for Beholder creature bank. So both `sounds/hota/dwellings/fangarm.wav` and `sounds/hota/creatureBanks/beholderSanctuary.wav` will not correctly describe this sound.

IvanSavenko:
Anyway, will probably merge this as it, before 1.3 release

krs0:
My intention behind HotA Mod changes is to have a complex mod that can be used as example for any other faction or complex mods. So any 'deviations' need to also apply for other mods. (One of them is SoD :) )

One thing I did not want to do was modify filenames. Because Factory will eventually come out and then it would help to keep same names, or at least to have a mapping like the one you did for maps.

I will look upon compacting some more subfolders after 1.3. But as feedback modding structure overall needs to be a little simplified.

Hota/config/hotaCove/creatures is not needed. We can specify a root folder property for a mod EG: Hota/cove/ and then have: HotA/config/creatures that resolves to HotA/config/HotA/cove/creatures/

IvanSavenko:
> My intention behind HotA Mod changes is to have a complex mod that can be used as example for any other faction or complex mods. So any 'deviations' need to also apply for other mods. (One of them is SoD :) )

Most of what I said also applies to SoD. Base game does not have multiple angels/blackDragons/etc either so even with plain directory structure there will be no file name conflicts. Especially if mod filesystems are isolated from each other

> One thing I did not want to do was modify filenames. Because Factory will eventually come out and then it would help to keep same names, or at least to have a mapping like the one you did for maps.

I actually have script that automatically imports all files from hota archive, with mapping similar to one in maps, so this won't be a problem

> Hota/config/hotaCove/creatures is not needed. We can specify a root folder property for a mod EG: Hota/cove/ and then have: HotA/config/creatures that resolves to HotA/config/HotA/cove/creatures/

True. If we look at path `hota/Mods/cove/Content/sounds/hota`:
- hota (first) - required, since this is top-level mod directory
- Mods/cove - can be merged into just "cove". However submods will have to be listed directly in mod.json. Not a bad change actually - we already had troubles with leftovers from old installs since some files are used even if not referenced anywhere, e.g. submods.
- Content - tricky one. We need to know where to scan for files that belong to this mod only (excluding submods). Removing it alongside with removing "Mods" directory would lead to a mess since submods will be present on the same directory level as data/sounds/sprites. However can be removed for submods IF we cut support for "submods of submods"
- sounds - can actually be (kind of) removed. Or at least - hardcoded requirement for it. Makes sense to have it for large mods but for mods that add 2-3 files not sure if complex directory structure is needed.
- hota (last) - can be removed as soon as mod filesystem isolation is done.

So with this system mods will have such directory layout:
```
ModName/
    mod.json <- main mod json, also contains list of submod names to look for
    Content/  <- files that belong to root/main mod
        config/
        sounds/
        sprites/
    Submod1/ <- files that belong to a submod
        mod.json <- mod.json of a submod
        config/
        sounds/
        sprites/
    Submod2/
        ... 
```
With this system depth of path to files in main mod will get cut by 1 and for submods - by 3.

And instead of submods of submods propose to use dependencies instead  to keep system simple.

Might actually be doable for 1.4...

PR #53, opened 2023-08-09
Seafaring Academy changes

krs0:
Limit is now 3 skills instead of 4 (like in HotA)
Reorganized json
image

### DIFF
--- a/hota/Mods/mapObjects/content/config/hotaMapObjects/seafaringAcademy.json
+++ b/hota/Mods/mapObjects/content/config/hotaMapObjects/seafaringAcademy.json
@@ -1,9 +1,14 @@
 {
 	"seafaringAcademy" : {
-		"handler":"market",
 		"name" : "Seafaring Academy",
+		"handler":"market",
 		"types" : {
 			"seafaringAcademy" : {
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Can somebody post a screenshot of this building from hota? It definitely should not like this.

@krs0 can you bump hota mod version, either here or in Junkman PR? So players will get update notification.

And we should probably limit any reorganization of mod structure / layout to avoid save breakage on 1.3. Can create 1.4 branch for such PR's

krs0:
Yes you are right. Upgrade is possible in HotA.
Having an empty slot is not a biggie... but not being able to upgrade is meh...

1st visit
image
3rd visit
image

krs0:
@IvanSavenko Tested in HotA:

Basically the building has 3 skills to offer. For each visiting hero looks to current hero skills and based on what is he missing he will get the next skill level.

- So no need to track anything on a per hero basis (just read the current skills the hero has and offer the according levels)

- A hero can visit as many times as he wants.

- A hero can update as many skills from the 3 as he wants.

IvanSavenko:
Thanks for research. Will update hota project with this info

PR #55, opened 2023-08-15
Homogenized HotA submod names and descriptions

krs0:
In VCMI Launcher the HotA submod names did not look homogenous.

HotA submods had a mix of capital and lowercase names
Some mods were named newfeatureX while others just featureX
Some names could express better what the mod is about


hota/Mods/hotAcampaignHeroesInRMG/mod.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name" : "Hota campaign Heroes in RMG",
+	"name" : "HotA campaign heroes in RMG",
```
IvanSavenko: It is not "in RMG" but general "in all single player maps".
Random maps, SoD maps
HotA maps are probably unaffected though.

PR #60, opened 2023-09-29
Removed unused LAND type from terrains

krs0:


DjWarmonger:
`LAND` property is not useless. VCMI also allows custom `WATER` terrains, which are not `LAND`.

krs0:
This you need to see with @IvanSavenko :)

hota/mod.json
```diff
@@ -6,10 +6,11 @@
 	"contact" : "http://forum.vcmi.eu/viewtopic.php?t=748",
 	"licenseName" : "Creative Commons Attribution-ShareAlike",
 	"licenseURL" : "http://creativecommons.org/licenses/by-sa/4.0/deed",
-	"version" : "1.6.12",
+	"version" : "1.6.13",
 	"downloadSize" : 114.9,
 	"changelog" : 
 	{
+		"1.6.13"   : [ "Removed unavailable LAND type from terrains." ],
```
Nordsoft91, line 13: I think this PR gives zero improvement (and changes) to the players, so why we should show them update available and ask to download 100+Mb?

krs0: Improvement is that there is no strange warning in client console.
Proposal on how to handle this?

Nordsoft91: But why do we have warning? Could you please paste the warning?

krs0: Was something like element type 0 not supported blah blah....

image

Nordsoft91: Understood, thanks. So, it was renamed. If there are more changes planned for hota mod prior 1.4 (probably hota quest gates?) release, I suggest do not bump version.

DjWarmonger: Okay, I see. We assume that the terrain is land if it's not water.

```
bool TerrainType::isLand() const
{
	return !isWater();
}

```
I can hardly imagine both LAND and WATER flags at same time, but who knows :P

krs0: What about `clouds`? with `sky ships`?

PR #63, opened 2023-10-02
Clearer magic shrines names

krs0:
All names look like:
image

Except level 1 that for some reason looks like:
image

(should be fixed later in a VCMI PR)


IvanSavenko:
Personally I would rather implement this on VCMI side. And not as part of name, but as additional part of description.

For example, your change will also rename these objects in map editor (and maybe in some other places)

krs0:
Yes, I thaught that as well, that these should be part of an additional field description.

I will close PR and open a generic one on VCMI side

PR #85, opened 2024-01-04
Add Ben Yan's Factory Port Prototype

LokiLaufeyjarson1712:


by003:
Does someone continue to work for other 12 heroes?

LokiLaufeyjarson1712:
@by003 I'll do, yeah

by003:
Thanks. 
@dydzio0614 Does it easy to develop some new abilities for factory? If responsed I will put my proposals here.

by003:
@LokiLaufeyjarson1712 or @misiokles 
Please update crimson couatl's def model here.
[crimCouatl.zip](https://github.com/vcmi-mods/horn-of-the-abyss/files/13837521/crimCouatl.zip)

This creature has two different def files in different states.
COUATLR is normal state:
image
COUATLRM is in meditation, this state is invincible.  (Maybe we use a bonus "INVINCIBLE".)
image

These two files were not exacted from HotA 1.7, I created them myself.

Unfortunately, HotA 1.7 did not use a normal def file to show them, they use an unopenable file which cannot be seen in a regular def previewer. Thus the turn left/right animations are lacked. If @dydzio0614 can develop the transformation of one creature with different def models in the battlefield, these are good for use as well as automations.

Now only normal def file can be used in the game.

PR #94, opened 2024-03-09
JSON fix Spanish translation

macomarivas:


misiokles:
Please updated spanish part of mod.json :)

PR #95, opened 2024-03-18
Adding Cartographer to modded mapObjects, changed limiter to 10,000

thiscris:
I saw [this on the Hota project board](https://github.com/orgs/vcmi/projects/3/views/1?pane=issue&itemId=24270615) and decided to use this to get familiar with the code base.
Steps that I took:

- Took the base config/objects/cartographer.json from the main repo;
- Removed the index parameters (for each type and the main one)
- renamed from "cartograhper" to "core:cartographer"
- changed the limiter and the resources { gold } from 1000 to 10000
- Saved the file under hota/Mods/MapObjects/content
- Added the file to the list hota/Mods/MapObjects/mod.json

I still haven't updated the text message, because I can't find the string in the repo. 
Could it be that it is hardcoded with the media files from the base (GOG) installation?
If somebody guides me how to update the text, I would extend this pull request

cartographer

### DIFF
--- a/hota/Mods/gameBalance/content/config/hotaGameBalance/mapObjects.json
+++ b/hota/Mods/gameBalance/content/config/hotaGameBalance/mapObjects.json
@@ -1,0 +1,101 @@
+{
+	"core:cartographer" : {
+		"name": "Cartographer",
+		"handler": "configurable",
+		"base" : {
+			"sounds" : {
+				"visit" : ["LIGHTHOUSE"]
+			}
+		},
+		"types" : {
+			"cartographerWater" : {
+				"aiValue" : 5000,
+				"rmg" : {
+					"zoneLimit" : 1,
+					"value" : 5000,
+					"rarity" : 20
+				},
+				"compatibilityIdentifiers" : [ "water" ],
+				"visitMode" : "unlimited",
+				"canRefuse" : true,
+				"rewards" : [
+					{
+						"limiter" : { "resources" : { "gold" : 10000 } },
+						"message" : 25,
+						"resources" : {
+							"gold" : -10000
+						},
+						"revealTiles" : {
+							"water" : 1
+						}
+					}
+				],
+				"onEmptyMessage" : "{Cartographer}\r\n\r\n28",
+				"onSelectMessage" : "{Cartographer}\r\n\r\nA big show is taking place at the Colosseum today. You enter the arena and hold the audience in awe with your magical prowess. Insatiable hunger for knowledge is what makes a true mage, so you gladly accept the first prize, which is one the newest academical handbooks",
+				"onVisitedMessage": "{Cartographer}\r\n\r\nModerator of the Colosseum instantly recognises you but respectfully denies entrance to the competitor zone. You probably should rather be featuring as a talent judge here!"
+				
+			},
+			"cartographerLand" : {
+
+				"aiValue": 10000,
+				"rmg" : {
+					"zoneLimit" : 1,
+					"value" : 10000,
+					"rarity" : 2
+				},
+				"compatibilityIdentifiers" : [ "land" ],
+				"visitMode" : "unlimited",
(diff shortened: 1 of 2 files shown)


DjWarmonger:
> Could it be that it is hardcoded with the media files from the base (GOG) installation?

Unfortunately that's the case. In order to properly implement it, you'd have to add new translatable text to object config - check any other object, for instance colosseum. Then this will require translations in every language.

Also, this change should belong to hotaGameBalance submod.

Then, bump the version in `mod.json` so that mod can be updated automatically.

thiscris:
I am sorry, could you give me some more information about the texts?
The only colosseum example that I could find was the Colosseum of the Magi (a hota map object). There the texts are in the format:
```
"onSelectMessage" : "{Colosseum Of The Magi}\r\n\r\nA big show is taking place at the Colosseum today. You enter the arena and hold the audience in awe with your magical prowess. Insatiable hunger for knowledge is what makes a true mage, so you gladly accept the first prize, which is one the newest academical handbooks",
"onVisitedMessage": "{Colosseum Of The Magi}\r\n\r\nModerator of the Colosseum instantly recognises you but respectfully denies entrance to the competitor zone. You probably should rather be featuring as a talent judge here!",
```

Replacing these lines from the original cartographer json didn't work
```
"onEmptyMessage" : 28,
"onVisitedMessage" : 24
```
I tried with `{Cartographer}` , `{cartographer}` and `{core:cartographer}` with no results.
I am only modifying json files and not making a new build. This should be enough, right?



> Then this will require translations in every language.

Is it possible to see how the text looks like in different languages without installing multiple GOG versions?

> Also, this change should belong to hotaGameBalance submod.

~~I am sorry, can you expand on that? I couldn't find a separate repo or branch for "hotaGameBalance submod".~~ 
edit: I found the folder gameBalance and I redid my changes to be part of it.

thiscris:
Ok so in this commit I moved the cartographer update to the gameBalance hota submod as requested.
I bumped the version of the submod and added a changelog entry.

Important to point out:
the cartographer texts are just copies of the colloseum of the magi texts, they are there as placeholders until they start working.
Any advice why my current json does not overwrite the standard cartographer text is much appreciated.

IvanSavenko:
From what I see, actual text with cost comes from here:
```js
"message" : 25,
```
Replacing these lines with custom text should work.

thiscris:
Wow. That seems so logical.
And the "onEmptyMessage" is shown when you don't have enough gold, while the "onVisitedMessage" is for when you have already purchased a map. I updated all 3 in the latest commit.
I didn't use "{Cartographer}" in the beginning of the messages, because I don't know what the text inside curly brackets is supposed to do.

There is a bug in VCMI where the "onVisitedMessage" is not being displayed. Instead it shows the original purchase offer as if you didn't buy the map already. Should I open a separate bug for it?

Question about translations: I guess that if we leave things like this people in other languages will have the text with the wrong amount (1000 gold) in their respective messages, even though the image will show the correct 10000. Is there a way to find these texts in other languages without having multiple OH3 installations with different localization?

IvanSavenko:
While those names might not be perfect from modmaker point of view their internal meaning does makes sense:
- onEmptyMessage: this is for cases when none of rewards are applicable to visiting hero and player should see that there is nothing here for his hero to use (e.g. empty).
- onVisitedMessage: this is when object has been already visited before.

> I didn't use "{Cartographer}" in the beginning of the messages, because I don't know what the text inside curly brackets is supposed to do.

This text is used to show window title. `{Cartographer}\n` means "show word Cartographer in yellow color and add line break after it. Not 100% sure but vcmi might automatically add title using object name if it was not present in message text.

> Is there a way to find these texts in other languages without having multiple OH3 installations with different localization?

Unfortunately no. There is already similar problem with secondary skills changed by hota (with different descriptions). Something to improve from vcmi side of things once I have some time.

PR #114, opened 2024-08-23
Fix mini Hill Fort creature level 5 upgrade

MichalZr6:
Upgrade creatures of level 5 should not be available.


IvanSavenko:
Are you sure? Haven't checked it, but at least wiki says that 5th level can be upgraded:
https://heroes.thelazy.net/index.php/New_Hill_Fort

MichalZr6:
Checked with original HOTA. I saw that page you mention. It's also a bit confusing that they give the info 1 - 5 you can upgrade at double the cost and 5 - 7 you can't. So level 5 belongs to which group ? :)

MichalZr6:
I did re-check. It seems I am wrong.

PR #176, opened 2024-12-29
avoid overriding hd mod data

Laserlicht:
@misiokles

### DIFF
--- a/hota/Mods/heroes3DataPatch/mod.json
+++ b/hota/Mods/heroes3DataPatch/mod.json
@@ -10,6 +10,10 @@
 		"1.3.6"   : [ "Added all changed sounds" ]
 	},
 	"depends" : [ "hota.mapdecorations", "hota.highlandsterrain" ],
+	"conflicts" :
+	[
+		"hd_version"
+	],
 	"softDepends" :
 	[
 		"wake-of-gods.animatedobjects",
--- a/hota/Mods/newGraphics/mod.json
+++ b/hota/Mods/newGraphics/mod.json
@@ -11,6 +11,10 @@
 	[
 		"ab-bad-ending-assets"
 	],
+	"conflicts" :
+	[
+		"hd_version"
+	],
 	"chinese" :
 	{
 		"name" : "城镇和生物的新冒险地图图像",


Olegmods:
@Laserlicht maybe it'll be better if the HD mod will soft depend on HotA?

Laserlicht:
> @Laserlicht maybe it'll be better if the HD mod will soft depend on HotA?

Also possible. But prefer it that way. With `softDepend` it's invisible what happens in background. With `conflicts` it's clearly visible in launcher.

Olegmods:
> > @Laserlicht maybe it'll be better if the HD mod will soft depend on HotA?
> 
> Also possible. But prefer it that way. With `softDepend` it's invisible what happens in background. With `conflicts` it's clearly visible in launcher.

Well, a lot of people (myself included) will prefer HD mod to be compatible with HotA, to be fair.

misiokles:
From what I see, it's still compatibke with hota, only two graphical submods are excluded.

Laserlicht:
Yes. Only two submods with new graphics are marked as conflict (they will override hd graphics).

Basically it's playable without problems. Only thing is missing these new graphics.

Olegmods:
> From what I see, it's still compatible with hota, only two graphical submods are excluded.

>Basically it's playable without problems. Only thing is missing these new graphics.

The thing is that HotA adds some more things than HD Edition mod, for example, some portraits not present in HD were fixed.
But current implementation is sort of okay, but I suspect 'softDepends' in HD mod will be better.

Laserlicht:
> The thing is that HotA adds some more things than HD Edition mod, for example, some portraits not present in HD were fixed. But current implementation is sort of okay, but I suspect 'softDepends' in HD mod will be better.

Yes. But then we need full hd support here in HotA. :)

I think more people will asking: "Why is my game not in HD even if I installed the mod?" than "Why are some portraits looks minimal different than in HotA?"

And changing HD mod isn't that easy because there is no update process.

Olegmods:
> Yes. But then we need full hd support here in HotA. :)

Seriously, that's a good idea, but it needs a lot of work with upscaling stuff and doing things in HDE (HD Edition) style, so it's not easy obviously.

> I think more people will asking: "Why is my game not in HD even if I installed the mod?" than "Why are some portraits looks minimal different than in HotA?"

I meant the portraits that aren't HD even with HDE mod, like Trolls. If there will be soft dependency in HDE mod, Trolls will have their HotA portrait (not SoD one) and Pikeman will use HD portrait. But the only problem is Elementals' portraits, which are neutral in RoE (and HDE).

> And changing HD mod isn't that easy because there is no update process.

If that's the problem, then sure.

PR #188, opened 2025-01-04
Add json validation tool

IvanSavenko:
Automatically check for json syntax issues, based on similar check in vcmi CI.


Olegmods:
> Automatically check for json syntax issues, based on similar check in vcmi CI.

It will check only syntax stuff, am I right?

PR #193, opened 2025-01-07
Typo fix

Olegmods:


misiokles:
Merged, but no need to boost version for such tiny changes :)

Olegmods:
> Merged, but no need to boost version for such tiny changes :)

Ok. Didn't know that.

PR #195, opened 2025-01-19
DEF files upgraded to 1.7.2

GeorgeK1ng:
All available DEF files in mod were replaced with DEF files from 1.7.2 HotA.lod

DEF files with D32F Header were excluded, same as D32 files. These must be done different way.

These DEF files are in reality D32F files:
artifact.def
artifbon.def
bores.def
COUATL.def
CPRSMALL.def
OVSLOT.def
PSKIL32.def
RCOUATA.def
RCOUATL.def
resource.def
secsk32.def
secsk82.def
secskill.def
TBFaDw_1.def
TBFaUp_1.def
twcrport.def
un32.def
un44.def

H3DataPatch was updated with new files aswell. So now every existing DEF files are now same binaries as in HotA.lod from 1.7.2

In this case there are still some not updated DEFs in mod which are Cove related. For future updates is needed to modify our Cove to match original HotA filenames for easier update.

After this PR is needed to check again for difference between HotA.lod and our mod and detect every DEF/D32F files which needs to be jsoned (even as placeholder with "Object is not implemented yet") to try to bring 1.7.2 map support.
Of course Cove needs to be touched to make this easier in future updates.


misiokles:
Actually we don't need to use many of them. For example artifacts.def cantains all oh3 artifacts sprites. And HotA ones are duplicates of pngs. HotA should be treated as any other mods.

GeorgeK1ng:
Yes, depends of modded file content.

PR #231, opened 2025-05-24
Fix swamp mountains

DjWarmonger:
- Fix mountain blockmap to match hota editor
image

- Promote their biome type to actual "mountains" (previously they looked too small)


GeorgeK1ng:
Check please also 1.7.2 branch, if we need same or similar fix there.

PR #239, opened 2025-06-14
Fixes for HotA changed visuals

### LINKED ISSUE #237: Fix changed shooting visuals submod
That mod contains some things not from HotA:

1. Cloudy Liches makes liches non-AoE.
2. Friendly Fireball is unnecessary, since shooting on an empty hex is implemented by using G key AFAIK.

Olegmods:
Fixes #237 

Liches now use normal AoE attack instead of whatever it was in this submod.

Friendly fireball removed since it was a workaround and you can use Magog attack on an empty hex using G key, and it was a workaround.

I suspect that there is some other non-HotA content, but won't change it yet.


GeorgeK1ng:
Any other change here? I don't know if these changes are correct, so it's up to you.

Olegmods:
> Any other change here? I don't know if these changes are correct, so it's up to you.

Well, I'd rather delete the whole submod, since likely most features of that mod aren't from HotA.
The progress with this PR stopped mainly because I need a confirmation if some of the spell-like attacks behave here like they do in HotA.

Olegmods:
Checked all the stuff. Only these two (friendly fireball and small lich cloud) aren't in HotA.

GeorgeK1ng:
So remove cloudyLiches and friendlyFireball only? If yes, I will do it directly, can't resolve conflicts from browser right now.

misiokles:
It means that Expl_05.def should also be removed from mod, as unused.

GeorgeK1ng:
Ok, will do that later today when I'm back on PC.

PR #241, opened 2025-06-25
HotA 1.7.3

GeorgeK1ng:
- Complete update to VCMI 1.7 format
- Added maps and campaigns support for current 1.7.3 HotA release
- Campaigns are now accessible from main menu
- DEFs were updated to 1.7.3 base, there are still some missing files - will be added in near future
- D32 files were extracted as JSON + PNGs, so these HotA files are now updated too
- Game balance was splited to individual submods - needs to create starting armies from 1.7.3, currently there are only blank files with comment what needs to be changed
- Current Game balance require review if all changes are actual, feel free to contribute
- AirShipyard is still "under construction", needs support from engine
- AirShip not fully implemented, require special button to get off hero from AirShip


hota/Mods/changedShootingVisuals/content/config/friendlyFireball.json
```diff
@@ -52,9 +51,14 @@
...
 		},
+		//"targetCondition" : {
+		//	"noneOf" : {
+		//		"bonus.DIRECT_DAMAGE_IMMUNITY" : "normal"  // Replaced by SPELL_DAMAGE_REDUCTION in VCMI 1.4
+		//	}
+		//},
 		"targetCondition" : {
 			"noneOf" : {
-				"bonus.DIRECT_DAMAGE_IMMUNITY" : "normal"
+				"bonus.SPELL_DAMAGE_REDUCTION" : 100
```
Laserlicht, line 61: @GeorgeK1ng This change isn't valid...

I suggest you to use debug version of vcmi, that triggers asserts...

GeorgeK1ng: Any idea how it should be?
There is also this PR, which removes it, so maybe solution? https://github.com/vcmi-mods/horn-of-the-abyss/pull/239

Laserlicht: Not sure about this stuff.

But `"bonus.SPELL_DAMAGE_REDUCTION" : "normal"` at least do not trigger the assert.

XCOM-HUB: SPELL_DAMAGE_REDUCTION could be named SPELL_DAMAGE_PERCENTILE — just an idea.

PR #263, opened 2025-11-24
Fixed Beatrice portraits

Aph-ra:
No idea what was going on with her portraits before but they displayed oddly in game:
image

Now they display correctly:
image


GeorgeK1ng:
Prabably do same for 1.7 repo (if needed)?

PR #267, opened 2025-12-01
fix couatl map blinking

mrhaandi:
The `AVWcoat.def` contains more frames than `AVWcoat.json`. This makes Couatls blink on the adventure map (visible / invisible).

Reproduce: create a random map and watch some neutral Couatls on the adventure map.

How was the initial .json created?
Here is a proposition how to fix this issue by adding frames to .json with same names as in the .def file.


GeorgeK1ng:
What version of AVWcoat.def have you used? One from HotA 1.7.3 is just placeholder.

image

mrhaandi:
HotA Installed version: 1.7.722, yes `AVWcoat.def` is a placeholder, but it contains frame information:
frames
Observe that some frames such as `AVWctl01,wgd` are repeated to have idle time. When simply extracted this information is lost and needs to be recovered in json (I guess).

mrhaandi:
Rebased to current `vcmi-1.7` branch.

mrhaandi:
Why is there a dummy .def file in the first place? Just a .json animation specification suffices to load an animation (vcmi 1.7).
The current blinking bug is because of the mismatch between .def frame info (with duplicate frames) and .json data (no duplication after extraction).

PR #271, opened 2025-12-12
add war machine damage cap mechanics

mrhaandi:
To be merged after https://github.com/vcmi/vcmi/pull/6436.

Given the corresponding secondary skill a war machine will not take more then 40% of max health damage.


mrhaandi:
The underlying mechanics are now part of vcmi, so this can be merged.

PR #277, opened 2025-12-17
New Pathfinding submod

Karyoplasma:
HotA changed pathfinding to also reduce base tile movement cost for off-road tiles to 95/90/85, so that Pathfinding has an effect on neutral terrains. This submod reflects this change.


misiokles:
There was a discussion some time ago how to handle 1.7.3 changes - to add them as 'standard' part of HotA or make submod with all possible to port 1.7.3 changes. And There was no consensus :)
Can we wait with merging of this commit till January to see if there will new hota patch will another balance changes??

Issue #285, opened 2025-12-21
Create HotA's first aid effect


by003:
Add limiters for first aid (it is already existed, created by @mrhaandi )
```
"core:firstAid" : {
		"base" : {
			"effects" : {
				"damageCap" : {
					"type" : "DAMAGE_RECEIVED_CAP",
					"propagator" : "ARMY",
					"limiters" : [{
						"type" : "CREATURE_TYPE_LIMITER",
						"creature" : "firstAidTent"
					}],
					"valueType" : "INDEPENDENT_MIN"
				},
				"hitPointUp1" : {
					"type" : "MORE_DAMAGE_FROM_SPELL",
					"propagator" : "ARMY",
					"hidden": true
				}
			}
		},
		"basic" : {
			"effects" : {
				"damageCap" : { "val" : 40 },
				"hitPointUp1" : {
					"subtype" : "bless",
					"val" : 0,
					"hidden": true
				}
			}
		},
		"advanced" : {
			"effects" : {
				"damageCap" : { "val" : 40 },
				"hitPointUp1" : {
					"subtype" : "slow",
					"val" : 0,
					"hidden": true
				}
			}
		},
		"expert" : {
			"effects" : {
				"damageCap" : { "val" : 40 },
				"hitPointUp1" : {
					"subtype" : "prayer",
					"val" : 0,
					"hidden": true
				}
			}
		}
	},
```
Then create a new json here:
```
{
	"core:firstAidTent":
	{
		"bonuses" : {
			"health1" : {
				"type" : "STACK_HEALTH",
				"val" : 5,
				"valueType" : "PERCENT_TO_BASE",
				"limiters" : [
					{
						"type" : "HAS_ANOTHER_BONUS_LIMITER",
						"bonusType" : "MORE_DAMAGE_FROM_SPELL",
						"bonusSubtype": "spell.bless"
					}
				]
			},
			"health2" : {
				"type" : "STACK_HEALTH",
				"val" : 10,
				"valueType" : "PERCENT_TO_BASE",
				"limiters" : [
					{
						"type" : "HAS_ANOTHER_BONUS_LIMITER",
						"bonusType" : "MORE_DAMAGE_FROM_SPELL",
						"bonusSubtype": "spell.slow"
					}
				]
			},
			"health3" : {
				"type" : "STACK_HEALTH",
				"val" : 15,
				"valueType" : "PERCENT_TO_BASE",
				"limiters" : [
					{
[...]

PR #286, opened 2025-12-21
Add stack health bonus to first aid

Karyoplasma:
My proposition for #285 

As I stated in the issue thread, this is not perfect. The main issues are:
+ Bonus health is not removed when the tent dies or the hero does not have a tent at all
    + I simply dunno how to fix this, the latter seems doable in a hacky way, but the former is impossible with current VCMI
+ specialty scaling only works for heroes in "core". VCMI issue, Ivan is aware but needs time to properly fix it.

 It's just less complicated and hacky than that abomination that was suggested is all.


misiokles:
I'll look into it tommorrow. But specialty scalling to heroes from mods fix was merged yesterday'.
https://github.com/vcmi/vcmi/pull/6477

Olegmods:
> - Bonus health is not removed when the tent dies or the hero does not have a tent at all
>   - I simply dunno how to fix this, the latter seems doable in a hacky way, but the former is impossible with current VCMI

Just as an idea (idk if it'll work, but it should, theoretically), but try to make a spell that makes the unit to add the stack health bonus to the army, then add ENCHANTED bonus to the First Aid Tent with that spell.

Karyoplasma:
> I'll look into it tommorrow. But specialty scalling to heroes from mods fix was merged yesterday'. [vcmi/vcmi#6477](https://github.com/vcmi/vcmi/pull/6477)

Awesome! Time to update VCMI
 
> Just as an idea (idk if it'll work), but try to make a spell that makes the unit to add the stack health bonus to the army, then add ENCHANTED bonus to the First Aid Tent with that spell.

It would have the same issues that I tried to avoid by binding it directly to the skill, namely not working in banks and not showing the health bonus outside of battle. But the idea is pretty good.

Karyoplasma:
Okay, guys. It is now 6:40-ish AM and I am finally happy with the solution I came up with this last commit (sorry for the amend, but this was just a much better version).

I toyed around with a custom spell as @Olegmods suggested and came up empty, so here is what I did instead:
I made a custum bonusType `FIRST_AID_AVAILABLE` and gave the First Aid Tent artifact an effect that spreads this bonus on any non-siege unit in your army. The `stackhp` bonus in the First Aid skill now checks against this bonusType.

It fixes everything apart from that the bonus is not removed when the tent dies in battle and I sincerely believe this is not fixable in the current version.
+ Works in banks. Even tho the tent is not present, the artifact apparently is
+ Health bonus is displayed in the creature info in your hero's inventory screen
+ If you run around with your hero that has First Aid but you don't have a tent, you get nothing as you should
+ `"secondary" : "firstAid"` scales the health bonus properly (I tested)

I know some people dislike the bonus icons, so set it the bonusType to hidden before merging if it bothers too much. I'm going to sleep now :)

PR #290, opened 2025-12-23
Specialty changes from 1.7.2

Karyoplasma:
This is just an example of how I would structure it.

I think individual hero files are probably better because as you can see with Darkstorn, the changes can become quite large. Most files will contain one or two-liners tho. Internally, I just bunch all of the stuff into one huge JSON xD


misiokles:
Looks great!

misiokles:
Suggestion. Maybe this submod should be named Hota1.7.2 balance or similar as main submod is all with hota balances...

GeorgeK1ng:
I would like to avoid numbering in balance. Just keeping latest possible changes.

misiokles:
@dydzio0614 can you repeat tenth time prons and cons of excluding 1.7.2 balance?

Karyoplasma:
The reason is that some people hate the hota balance changes, I know that and it puts a mod like "gameBalance" is in an odd spot. I agree tho that the "gameBalance" mod should reflect the latest changes and that there should be a separate mod for old scaling that has all the initial changes to it.

Reverting all of what is already changed is borderline impossible at this point tho. What does the mod change that is integral to HotA tho? I haven't played HotA before 1.6.0, so I have no idea what was going on before that.

You could just clone the mod as it is right now and call it "old scaling". It literally only changed Deemer, Uland starting skills and gave Gem bless. That was all "gameBalance.heroes" did...

dydzio0614:
lets just change things to match hota 1.7.2 for now... in general changes from 1.7.2 as a whole are not liked by some people - war machines changes, easier dragon dwelling building in rampart/dungeon, secondary skills changes etc.

Karyoplasma:
The mod.json and the folders contain skeletons for the heroes I didn't change yet. From memory those are:

Astra - no idea how the cure specialty worked before and what the change is
Uland - i kept the changes that were previously in heroes.json, but he is also a cure specialist
Zilare - no idea how Forgetfulness works mechanically
Coronius - oh boy this one is gonna be a hassle. We'd better ask for a VCMI internal fix for that lol. It might be possible somehow, but I don't see a way
Frederick - has `"creature" : "automaton"` which is probably a placeholder, but I updated the text anyway
Dargem - I did not test this, but reading through airShield, this should work.

The ones without comments in the mod.json are tested changes which is the majority. They were part of my private mod where I fixed them.

PR #300, opened 2025-12-28
Made Elemental summon spells non-exclusive

Karyoplasma:
1.7.0
[+] Lifted the ban on summoning different types of elementals in the same combat

Also cleaned up the existing adjustments to make it easier to read. No functional change besides making Summoning X Elemental non-exclusive.

### DIFF
--- a/hota/Mods/gameBalance/mods/spells/content/config/hotaGameBalance/spells.json
+++ b/hota/Mods/gameBalance/mods/spells/content/config/hotaGameBalance/spells.json
@@ -159,13 +159,18 @@
 	"core:fireElemental" : {
 		"levels" : {
 			"base" : {
-				"power" : 70,
 				"battleEffects" : {
 					"summon": {
 						"exclusive" : false,
 						"summonByHealth" : true
 					}
 				}
+			},
+			"none" : {
+				"power" : 70
+			},
+			"basic" : {
+				"power" : 70
 			},
 			"advanced" : {
 				"power" : 88
@@ -178,13 +183,18 @@
 	"core:earthElemental" : {
 		"levels" : {
 			"base" : {
-				"power" : 80,
 				"battleEffects" : {
 					"summon" : {
 						"exclusive" : false,
 						"summonByHealth" : true
 					}
 				}
+			},
+			"none" : {
+				"power" : 80
+			},
+			"basic" : {
+				"power" : 80
 			},
 			"advanced" : {
 				"power" : 100
--- a/hota/Mods/gameBalance/mods/spells/content/config/hotaGameBalance/spells.json
+++ b/hota/Mods/gameBalance/mods/spells/content/config/hotaGameBalance/spells.json
@@ -5,9 +5,6 @@
 			"base" : {
 				"effects" : {
 					"spellDamageReduction" : {
-						"type" : "SPELL_DAMAGE_REDUCTION",
-						"subtype" : "air",
-						"duration" : "N_TURNS",
 						"val" : 50
 					}
 				}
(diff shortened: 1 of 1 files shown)


Karyoplasma:
please wait before merging, i have an idea to get the exact scaling of Earth/Fire eles (2/2/2.5/3)*SP

GeorgeK1ng:
> please wait before merging, i have an idea to get the exact scaling of Earth/Fire eles (2/2/2.5/3)*SP

Okey regarding to exclusivity. We should then check Its compatibilty with "new summoning spells" mod

Karyoplasma:
Earth and Fire Elementals could use `"summonByHealth"` with spell power multiplied by 40 (Earth) and 35 (Fire). It's still not perfect because 35*2.5 is not an integer, but better than before (was same summon amount on advanced and expert). In my test, a hero with 20 spell power and advanced Fire Magic and expert Earth Magic summoned 50 Fire Elementals and 60 Earth Elementals, which is correct.

It only modifies the core versions, so it doesn't break any mods that add new summon spells as they would have their on spell power.

GeorgeK1ng:
That mod modify core ones too. Will take care about it then.
Example - 
 https://github.com/vcmi-mods/new-summon-spells/blob/vcmi-1.7/new-summon-spells/mods/onlyOneSummon/content/config/spells/summonElementals.json

Karyoplasma:
It wouldn't conflict. It would just reset the exclusivity flag (and use HotA's scaling). But it would use HotA's scaling before as well, just the more inaccurate version.

GeorgeK1ng:
There will be 100% warnings about mods load orders in console.

Karyoplasma:
Of course. But what does that even mean? It would still work as it is supposed to. The exclusivity change alone would trigger a warning in the console since one mod sets it and another one resets it.

GeorgeK1ng:
No it doesnt work like that, it can conflicts each other. Will show you then console about that when Im back on PC.

Karyoplasma:
this is the config of advanced level of Earth Elemental when using the OnlyOneSummon mod:
```
"advanced" : {
			 // core
			"battleEffects" : {
				 // core
				"summon" : {
					 // new-summon-spells.onlyonesummon
					"exclusive" : true,
					 // new-summon-spells.onlyonesummon
					"id" : "earthElemental",
					 // new-summon-spells.onlyonesummon
					"permanent" : false,
					 // hota.gamebalance.spells
					"summonByHealth" : true,
					 // new-summon-spells.onlyonesummon
					"type" : "core:summon"
				}
			},
			 // core
			"cost" : 20,
			 // new-summon-spells.onlyonesummon
			"description" : "{Advanced Summon Earth Elemental}\r\n\r\nAllows you to summon earth elementals.\r\n\r\nOnce cast, no other creatures can be summoned.",
			 // hota.gamebalance.spells
			"power" : 100,
			 // core
			"range" : "X"
		}
```
Looks fine. But I now realize what you meant with it could conflict because the OnlyOneSummon could have overwritten the health flag with the base setting, which is false.

Side note, the mod.json for OnlyOneSummon.summonHotaCreatures is broken because it doesn't depend on hota.cove, so the loader doesn't find Nymphs or Stormbirds. I needed to add a hard dependency before it would load. Maybe that was a VCMI 1.7 change, idk, but as is, the mod does not load at all with HotA.

misiokles:
long story short - should I merge this commit now? Cause I'm lost again:)

Karyoplasma:
I'd wait for a reply from @GeorgeK1ng 

In my eyes it's fine, tho

GeorgeK1ng:
@misiokles Merging, just check how it goes with mentioned spells mod. We have just few spells mods and players have them enabled :)

PR #304, opened 2026-01-01
Add Mysterious Island v1.4n

GeorgeK1ng:
1) Added Mysterious Island v1.4n, which is latest vesion for HotA 1.7.3. So most wanted community map joined our mod, because downloaded one from maps4heroes is now for HotA 1.8 and will not work in VCMI
2) Added initial AI translation for this map for Czech

No version bump for now!


IvanSavenko:
Why not add that map as separate mod?
For visibility & to make it clear that map is not part of hota?

GeorgeK1ng:
Will separate it little bit later with few more famous maps to "Community maps" submod, but enabled by default. This specific map is searched a lot.. Right now I wanted to archive last playable one until HotA 1.8 map support.

PR #308, opened 2026-01-02
Bulwark Creatures port (beginning)

Karyoplasma:
This is the skeleton of the creature port from Bulwark. No mod.json, because I was lazy.
Missing stuff:
+ sound files for unit sounds
+ some creature abilities
    + runes for upped yetis
    + war mammoth defense bonus on defensive stance ("DEFENSIVE_STANCE" does not accept percentages, will make VCMI issue later)
    + arkali aura i simply dunno how to do without something that properly simulates an aura with a range (will make VCMI issue later and beg for tech)
    + freezing shot from greater shamans (i dunno how it works, didn't play enough)
    + jotunn/jotunn warlords crash vcmi when they cast teleport


GeorgeK1ng:
@by003 Let me know how to coordinate. Im sure you wanna to take care about town porting, so creatures helps mapping should be helpful!

Karyoplasma:
I copied the structure from the cove submod if that helps. I found it easier to work with as the bulwark version.

by003:
I think it is necessary to build the file and make it running in game. Some of mechanics or creature abilities can be left blank temporarily and fixed later. Meanwhile, if it's not a very complex mechanism, the development team can be asked to assist in writing or correcting it.

Karyoplasma:
found some more pictures in the hota-180 archive from #305 (ok, you got me, i didn't look out of laziness, they were there from the start)
also added the large icons for units
Edit:
found the projectile defs in the archive (defpreview suggests that they fit, it's a spear for the elves and some icy swirly thing for the shamans)

This is now complete aside from sounds and some non-working abilities. I posted a preview with pics and a small fight in the vcmi discord modding channel.

by003:
After analysis, I found that several configurations might be missing:

1. Configurable resistance aura, which should at least be configurable to resistance (like dwarf) or damage reduced (like golems), and the limiter should be configurable as opposite.

2. Configurable defense bonus, which should be configurable as a percentage.

3. Freeze animation effect. Similar as petrification and bloodlust.

Other bugs that need fixing:

Mainly spells like teleport and sacrifice, which have a two-stage casting process that creature units cannot control.

by003:
**For Runes skill:**
Only heroes with **scholar** ability can learn runes; this seems to be definable.

Also: Bulwark heroes can learn runes without scholar; could this be defined through alignment?

However, one point: the **'allOf'** in the limiter seems ineffective.

Regarding skill effects: Rune levels can be stacked for attacks, being hit, and defending.

First:
1. The target of a spell cast after an attack can only be nearby units. Although Location can be used to configure this, it must be within a range; it cannot only affect the caster themselves. Two years ago, I proposed a range: 's' to represent casting a spell on oneself.

2. Casting a spell after being hit can be configured using 'spell before attack,' which affects the caster but not when countered.

3. There is currently no option to cast a spell while defending.

4. Even if stacking is possible, only attack, defense, and speed can be stacked simultaneously, not attack first, then defense, then speed.

Karyoplasma:
Runes requires new tech. I didn't even know that you need scholar to learn it, the sheer functionality is impossible to recreate:
SPELL_BEFORE_ATTACK only works if your creature actually attacks or retaliates, if you are out of retaliations, it won't proc.
You need to determine which effects should be added. It's stacks +1 atk, +1 def, +1 speed in that order, up to a cap of 9 stacks total (3 per stat). So if you get attacked while having 8 stacks of runes, you should only get +1 speed even tho getting hit increases your rune stacks by 2, so you need to keep track of where in the chain you are.
Yeti Runemasters ability basically gives them Expert Runes. If the hero also has Expert Runes, all the effects are doubled, but the maximum stacks stay at 9.
I don't see a way to hack that in.

> Configurable resistance aura, which should at least be configurable to resistance (like dwarf) or damage reduced (like golems), and the limiter should be configurable as opposite.

Yes, it should work like Unicorn aura, so moving with the unit, affecting the adjacent hexes. Would be cool if we could set the range and what bonus it should apply when near. Issue is that the Unicorn aura is fixed in giving chance to resist and you cannot change that :(

>Configurable defense bonus, which should be configurable as a percentage.

There is DEFENSIVE_STANCE bonus which almost works, but it only works for flat values. I tried to make the bonus give 0 extra flat defense and then give the War Mammoth a second ability that boosts their defense by 80% limited to DEFENSIVE_STANCE, but that bonus is always active. As a placeholder, we could just give War Mammoth a DEFENSIVE_STANCE with a flat boost of their defense value.

> Freeze animation effect. Similar as petrification and bloodlust.

If the freeze animation has a def, Freezing Attack could be added with SPELL_AFTER_ATTACK then, we just need to know the proc chance and can clone paralyze with the appropriate animation (couldn't find the def tho). It it's just a colormuxer, we could add that, but it would require to load the entire battleEffects.json, including the effects of the base game, since it's not merged when loaded, so we cannot just add our effect to it.

Addendum:
The Jotunn aura limiter is super ugly, because it requires a hard dependency on cove and factory since you need to exclude teleporters. It would be good if you could filter for bonusSubtype to only apply the bonus to movementFlying, but it doesn't work. bonusSubtype.movementFlying exists already, you just cannot filter for it.
It would also be cool if the "stacking" field would work in absolutes. You have to give Jotunn Warlords 2 abilities currently: 1 that doesn't stack with Jotunns and one that does, both lowering the speed by 1. if you just set "stacking" with a bonus of -2, it will keep Jotunn's bonus when both are on the field because -1 is bigger than -2 (counter-intuitive, since -2 is the larger effect). The same is true for devils/arch devils and their -luck (that one is still bugged in gameBalance btw, I didn't think about that interaction when I "fixed" it).

GeorgeK1ng:
Will PR sounds now to fullfill these gaps.

by003:
> Will PR sounds now to fullfill these gaps.

I can make some substitute plans to fill the gaps temporarily. Similar like factory.

What did you think?

GeorgeK1ng:
> > Will PR sounds now to fullfill these gaps.
> 
> I can make some substitute plans to fill the gaps temporarily. Similar like factory.
> 
> What did you think?

Sounds for creatures are already done now.

PR #315, opened 2026-01-04
Siege

Karyoplasma:
Missing:
+ drw1 (gate opening) and drwc (the ropes/chains of the gate)
+ mlip (bank of the moat, not sure what that even is)
+ wa13, wa33, wa43, wa63 (the completely destroyed wall parts)
+ arrow tower images for the buttons

x value of keep creature needs adjustment.


Karyoplasma:
I'll send a better version (easy fixes)

PR #316, opened 2026-01-04
Siege battle

Karyoplasma:
@by003 assembled it, looks good, the only nitpick I have is that the creature on the keep looks a bit misplaced on the x-value (I've already adjusted it but it's still not perfect).

Icons for arrow towers are still using the borrowed ones from Cove, but other than that it's complete.


by003:
Fine. I can make new arrow tower icon later.

PR #318, opened 2026-01-04
Fix for Cove's bugged Puzzle Map

Karyoplasma:
The puzzle tile opening index was off by 1, leading to the first tile always disappearing without having any obelisks visited.


C3pa:
I guess such issues could be avoided if VCMI logged an error message when loading the definition for a puzzle map with an index of 0.

Karyoplasma:
@C3pa 
Yeah, no error or warning was shown. Also the docs are misleading in this case, too. This is the example:
```
"pieces" :
        [
            {
                // Position of image on screen
                "x" : 0
                "y" : 0

                //indicates order in which this image will be opened
                "index" : 0 
            },
            ...
        ]
```

C3pa:
I guess it would be wise to open an issue on the VCMI repo, mentioning both the mistake in the docs and a request for an error to be logged.

PR #357, opened 2026-01-12
Bonus icons

Karyoplasma:
Added the bonus icons from @by003 in #305 

War Mammoth icon does not show when put on the defense bonus for some godforsaken reason, so I gave it a bonus that does nothing to show the icon.

Removed the comment on Jotunn/Warlords KING_3 since I've confirmed it is indeed KING3

Fixed Rion's starting army in gamebalance/startingsArmies (why is there an extra s in the mod name btw?). He had no First Aid Tent.


by003:
I'm not sure why Jotunn has 'KING_3' bonus, in my opinion, the rules should be:
beast like creature: KING_1
human like creature: KING_2
human like and mind immunity: KING_3

PR #362, opened 2026-01-13
Fix Hermit's Shack

Karyoplasma:
Noticed this issue in VCMI: https://github.com/vcmi/vcmi/issues/6722 and went to investigate and yes, Hermit's shack was fundamentally broken and potentially rewarded multiple levels for a secondary and could also pick an Expert secondary skill to upgrade (which did nothing and marked the object as visited).

Fixes in this PR:
+ better object description
+ removed `"onSelectMessage"` (unused - only used when object is set to `"playerSelect"`)
+ fixed `"onEmptyMessage"` (used to be the same as `"onVisitedMessage"`)
+ fixed limiters and bonuses for all rewards

Now it behaves as it should: It gives you a random secondary skill upgrade (1 level, never multiple) for either a Basic or an Advanced secondary skill. If your hero has no Basic or Advanced secondary skill, it will display the `"onEmptyMessage"`, telling you that you cannot learn anything here at the moment. This does not mark the object as visited, so you can return to it later.

### DIFF
--- a/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/hermitsShack.json
+++ b/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/hermitsShack.json
@@ -5,7 +5,7 @@
 		"types" : {
 			"hermitsShack" : {
 				"name" : "Hermit's Shack",
-				"description" : "(Improves a random secondary skill)",
+				"description" : "(Random secondary skill upgrade once per hero)\n",
 				"rmg" : {
 					"value" : 1500,
 					"rarity" : 80,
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
Nice, now we need to add new skills using append compatibility submods and it's done.

Karyoplasma:
@GeorgeK1ng 
I can add Interference and Runes, but one has to use `"append"` and the other `"appendItems"` because it works in a weird way.
If you soft-depend, one append will just override the other and/or produce extremely weird results, depending on which version you use.

Other mods that add secondaries will face the same issue and have to keep track of which mod adds what in which way. Or we have a compatibility submod for every possible combination of mods which is laughably unfeasible. Every skill you add to the reward requires a hard-dependency.

GeorgeK1ng:
Yes, just adding latest to list, same as Market of Time. Each should be single one, one from Interference submod and one from Bulwark. Both should use only **append**

Karyoplasma:
@GeorgeK1ng 
That's not how `"append"` works currently. It will merge the appends before manipulating the vector, so it would create only 1 additional, nonsensical reward. Here is what happens when 2 mods use `"append"` to add a reward:
[hota.highlandsterrain_objects.hermitsShack.json](https://github.com/user-attachments/files/24600372/hota.highlandsterrain_objects.hermitsShack.json)

This is why you have to use `"append"` for one of them and `"appendItems"` for the other. All other mods that want to add to this list will fuck it up.

PR #363, opened 2026-01-14
Fix highlands objects again and fix launcher warnings

Karyoplasma:
This PR fixes overlooked issues in the hota.highlandsTerrain submod. It also fixes the launcher warnings about removing empty object sets.

hota.highlandsTerrain:
Objects:
+ fixed `"visitableFrom"` field of Junkman, Hermit's Shack, Mineral Spring and Obelisk
+ fixed mask of Mineral Spring, a spruce (avlhpn12) and Highlands Abandoned Mine 
+ added a Trading Post for highlands terrain (for some reason, RMG wouldn't spawn it otherwise, now it does)
+ Renamed the waterfalls from "Mountain" to "Waterfall"
+ removed Gazebo missing sound comment. There is no sound missing, Gazebo doesn't have an ambient sound and the interact sound was correct.

Biomes:
+ made Waterfalls "other" instead of "mountain", they look weird as mountains
+ removed "puddles" and "highlandHole" biomes since they had no objects with blocking tiles, so they were removed by the launcher with a warning
    + They shouldn't be defined as statics because statics are assumed to have blocking tiles, but I have no better idea for them
    + I do not know how to make RMG spawn non-blocking decorations :(

What is still missing is that the highland version of the churchyard is picked for highlands terrain by the RMG. It never is for me, it only ever spawns the normal version. The definition files are correct, so I have no idea what is wrong, it works for other terrains and objects.

hota.wastelandTerrain:
+ removed the "limestonePuddles" biome, empty set, no blocking objects

hota.mapDecorations:
+ removed the "kelp" biome, empty set, no blocking objects

This fixes all launcher warnings about empty object sets. The one remaining is a mistake in core's biomes.json.

### DIFF
--- a/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/gazebo.json
+++ b/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/gazebo.json
@@ -14,7 +14,7 @@
 				"sounds" : {
 					"visit" : [
 						"GAZEBO"
-					] //TODO: missing sound!
+					]
 				},
 				"rewards" : [
 					{
--- a/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/hermitsShack.json
+++ b/hota/Mods/highlandsTerrain/content/config/hota/highlands/hotaInteractive/hermitsShack.json
@@ -37,7 +37,7 @@
 						],
 						"visitableFrom" : [
 							"---",
-							"-++",
+							"++-",
 							"+++"
 						],
 						"allowedTerrains" : [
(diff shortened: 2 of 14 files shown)


GeorgeK1ng:
This: "I do not know how to make RMG spawn non-blocking decorations :(" is currently unsupported by engine.

misiokles:
How is your opinion about Wastaland's Grave object? Exact mechanics is of course not implemented, but.... shoudn't this object be refusable?

Karyoplasma:
> How is your opinion about Wastaland's Grave object? Exact mechanics is of course not implemented, but.... shoudn't this object be refusable?

Funny, I did some work on Grave to adjust the reward from an earlier patch (1.7.0) without reading your message (#365). But I agree, it should not just take your movement points like a gotcha. If you have a good idea for a message, go ahead. Sadly you cannot make a message that is shown before the rewards are rolled, so you will see the rewards before accepting/declining.

Edit: I changed it to refusable and gave it the shovel digging sound on removal. I've adjusted the message a bit for this.

PR #367, opened 2026-01-15
Fix terrain exclusive templates

Karyoplasma:
This PR fixes objects or object types introduced in HotA that have terrain variations in their templates, for example Churchyard, Black Tower or Experimental Shop. RMG would never spawn the terrain variation because the default version was allowed on all terrains. I don't know why that is a problem for newly introduced types as with core object types, you can just add the template (see learningStone in mapDecorations for example) and it'll work...

Going by object:
+ Black Tower
    + fixed a typo in the animation that made the normal variation spawn with the grass/swamp sprite
    + fixed grass, swamp and snow versions of it never spawning
+ Churchyard
    + fixed highlands and snow versions never spawning
+ Experimental Shop
    + fixed snow version never spawning
+ Ivory Tower
    + fixed sand variation spawning everywhere and snow version never spawning
    + corrected the rarity
    + made it default banned like in HotA
+ Wagon
   + Fixed the HotA version spawning everywhere (should only spawn on rough terrain)

Changes semi-related to map objects:
I didn't like how the wasteland and sand zones looked biome-wise, so I changed it a bit. I made only the skulls (small blockers) spawn as "animal" in wasteland and sand and made a new biome with type "other" for both wasteland and sand for the bigger skeletons.

### DIFF
--- a/hota/Mods/mapDecorations/content/config/hotaMapDecorations/vanillaInteractive/wagon.json
+++ b/hota/Mods/mapDecorations/content/config/hotaMapDecorations/vanillaInteractive/wagon.json
@@ -13,6 +13,9 @@
 							"---",
 							"-++",
 							"+++"
+						],
+						"allowedTerrains" : [
+							"rough"
 						]
 					}
 				}
--- a/hota/Mods/mapObjects/content/config/hotaMapObjects/blackTower.json
+++ b/hota/Mods/mapObjects/content/config/hotaMapObjects/blackTower.json
@@ -12,7 +12,7 @@
 				"guardsLayout" : "creatureBankBlackTower",
 				"templates" : {
 					"normal" : {
-						"animation" : "hota/banks/avblktgr.def",
+						"animation" : "hota/banks/avblktwr.def",
 						"visitableFrom" : [
 							"---",
 							"-+-",
@@ -23,6 +23,15 @@
 							"VVVV",
 							"VVVV",
 							"VVAV"
+						],
+						"allowedTerrains" : [
+							"dirt",
+							"sand",
+							"rough",
+							"subterra",
+							"lava",
+							"hota.wastelandTerrain:wasteland",
+							"hota.highlandsTerrain:highlands"
 						]
 					},
 					"grass" : {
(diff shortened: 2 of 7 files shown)


misiokles:
Objects not spawning could be a bug. It worked before. It's no use to list all available allowed terrains - what about additional terrains from other mods? I must check it closer.

GeorgeK1ng:
Regarding to Ivory Tower - made it default banned like in HotA. This should be done on Game balance objects submod instead, which is already there? https://github.com/vcmi-mods/horn-of-the-abyss/blob/hota-1.8.0/hota/Mods/gameBalance/mods/objects/mods/rmgBan/content/config/hotaGameBalance/objectsRmgValues.json
It's enabled by default and bring ability to keep generate it on Random Map when disabled.

GeorgeK1ng:
> Objects not spawning could be a bug. It worked before. It's no use to list all available allowed terrains - what about additional terrains from other mods? I must check it closer.

Maybe using **noneOf** will be better?

Karyoplasma:
> Regarding to Ivory Tower - made it default banned like in HotA. This should be done on Game balance objects submod instead, which is already there? https://github.com/vcmi-mods/horn-of-the-abyss/blob/hota-1.8.0/hota/Mods/gameBalance/mods/objects/mods/rmgBan/content/config/hotaGameBalance/objectsRmgValues.json 

Setting `"zoneLimit`" to 0 is bad, then the RMG will pick them, see they cannot be placed and it has to reroll the object (gives out warnings). Better just null the rmg? 

> It's enabled by default and bring ability to keep generate it on Random Map when disabled.

I don't understand this, it's a HotA object, so it'll never generate on other maps unless specifically enabled.

> Maybe using noneOf will be better?

Can you? Every time I tried, launcher complained about critical issues in the mod:
```
[2026-Jan-15 15:57:44.882048] ERROR [initialize] mod - Failed to resolve identifier 'noneOf' of type 'terrain' from mod 'hota.mapobjects'
[2026-Jan-15 15:57:44.883048] ERROR [initialize] mod - Identifier 'noneOf' of type 'terrain' does not exists in any loaded mod!
```

GeorgeK1ng:
Seems it's not implemented usage of **noneOf** in allowedTerrains.

Yes currently we can make them null or 0 0. Why not allow various objects for RMG when disabled that submod? 

Btw, objectsRefusable in game balance require message update for rewards.

Karyoplasma:
We confirmed this is a bug. RMG currently considers null in allowedTerrains to be more specific than a template with 1 allowedTerrain. This PR is not necessary as a whole. I will send one with stuff that I actually fixed instead.

PR #372, opened 2026-01-16
Black tower guard layout fix

Karyoplasma:
Black Tower battle is a regular battle, no fancy shenanigans.
I've removed the `"creatureBankBlackTower"` layout definition from the mod.json as it unnecessary (and it was wrong if you showed up with less than 7 units anyway).

Also fixed the visitableFrom mask.


misiokles:
I don't remember correctly, but Black Tower Layout was made to have obstacles and war machines on the battlefield, like in HotA

Karyoplasma:
Probably from a time when "default" guard layout didn't exist.

PR #380, opened 2026-01-17
Wasteland and Highlands biomes improve

Karyoplasma:
I wasn't satisfied with how often the large bones would spawn in wasteland zones, so I improved it a little. I've also fixed a wrong animation and added a missing tree.

Fixes:
+ Added a missing tree
+ Corrected a tree which was set to the wrong def
+ Used a forgotten cactus (was defined but unused)

Changes:
+ cleaned up the wasteLandTree biome, it had more templates than were defined in trees. i added all trees from "Wasteland Objects" in the hota map editor into this
+ added a dead looking tree that is not classed as dead vegetation into the wastelandDeadVegetations biome
+ made all rocks spawn in wastelandRock and added an identical wastelandStone set as "other" to compete with the bones
+ made the shrubs object type "other" and removed the decorative shrubs that are never spawned
+ split up the cacti in desert looking ones as plant and more verdant looking ones (like the flowering barrel cacti) as other
+ added the forgotten cactus into the verdant group

RMG still seems to love the wastingBones biome, but now there is at least a chance it isn't picked.

Edit: 
Couldn't sleep, so I went over the Highlands biomes again as well and corrected some "mistakes":
+ added a missing flower (vanilla flower that can also spawn on highlands)
+ added the missing flower to plant biome
+ re-classified a flower from flowers into shrubs (HotA calls it a shrub)
+ moved newly classed shrub from flowers biome to shrubs biome
+ made shrubs biome object type "other" for a more florally Highlands look

In total: same amount of flowers, 1 more shrub


GeorgeK1ng:
avltro13.def should be then removed from hota\Mods\mapDecorations\content\sprites\hota\trees to don't duplicate files.

Karyoplasma:
mapdecorations also uses it to make a new dirt tree biome tho. So what could maybe be done is remove the wasteland version in wasteland, make the mapdecorations version also spawn on wasteland and use their template for it.

Or would it be better to define it in wastelandTerrain and use it in mapdecorations? Optimally, biomes should ignore placement rules, so you don't have to redefine everything 17 times because x tree doesn't spawn on z terrain...

misiokles:
Map decorations mod was made before 'biomes' feature was added - so yeah, HotA mod needs refactoring. If you have a bigger picture how to do this nicely, so why not? Even more - it's better to do it now than later, because official 1.8 port will most likely break saves.

Karyoplasma:
I've opened an issue that would really help solving lots of issues: https://github.com/vcmi/vcmi/issues/6745

If biomes simply ignored the terrain rules of the given templates and stuck to the rules of, you know, the actual biome, we could just use already defined objects without having to worry about redefining each tree 17 times because flower13 does only spawn on grass while I want to use it on highlands...

If you look at [this](https://github.com/vcmi-mods/horn-of-the-abyss/blob/hota-1.8.0/hota/Mods/wastelandTerrain/content/config/hota/wasteland/vanillaStatic/trees.json), just count how many vanilla objects have to be redefined... You cannot even append that field, and all of them show up as duplicates in the map editor's object browser which is making it even less usable than it already is...

GeorgeK1ng:
> Map decorations mod was made before 'biomes' feature was added - so yeah, HotA mod needs refactoring. If you have a bigger picture how to do this nicely, so why not? Even more - it's better to do it now than later, because official 1.8 port will most likely break saves.

Exactly, saves are now 100% not compatible. So any bigger change can ve done now or next VCMI major release.

PR #387, opened 2026-01-18
Pre-configure support for Sea Barrel and Grave (I think)

Karyoplasma:
I gotta be honest here, I have no clue what a "mapDice" is and it also isn't explained anywhere, so I randomly plopped in some numbers until it seemed to work. Left it out of Grave entirely (not sure if that is a mistake either), but my testing said it's fine.

How I tested:
[bullshit.zip](https://github.com/user-attachments/files/24696806/bullshit.zip)
This map is called "zzzzzzgraveTest" ingame and contains a preconfigured Grave (upper one always gives 15 Gold and Boots of Speed) and a normal one. It also contains several sea barrels: top-left always gives some 50ish sulfur, bottom-left is always empty, others are random.

I restarted this map several times and the fixed rewards seem fixed and the random rewards seem random, so I guess that worked? Sea Barrel also logs the same warning as Wagon and Campfire, so that is reassuring.

About Ancient Lamp... yeah, that building doesn't even properly function, so I have not the slightest idea how to do this.

### DIFF
--- a/hota/Mods/mapObjects/content/config/hotaMapObjects/seaBarrel.json
+++ b/hota/Mods/mapObjects/content/config/hotaMapObjects/seaBarrel.json
@@ -25,11 +25,25 @@
 				},
 				"blockedVisitable" : true,
 				"removable" : true,
+				"variables" : {
+					"resource" : {
+						"gainedResource" : { 
+							"anyOf" : [ "mercury", "gems", "sulfur", "crystal" ]
+						}
+					},
+					"number" : {
+						"gainedAmount" : {
+							"min" : 3,
+							"max" : 6
+						}
+					}
+				},
 				"rewards" : [
 					{
 						"appearChance" : {
 							"max" : 20
 						},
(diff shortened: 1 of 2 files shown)


GeorgeK1ng:
Ancient Lamp should work just fine, just require this new config?

Karyoplasma:
> Ancient Lamp should work just fine, just require this new config?

How do you multiply the gold limiter by the amount of creatures you purchase exactly? The building is functionally broken.

GeorgeK1ng:
Here is updated docs https://github.com/vcmi/vcmi/pull/6753

Karyoplasma:
That does not help at all. You do not "hire" them, you just get them as a reward and lose gold. Limited by the gold you currently have, then it goes away even if you just hired one. It is not a dwelling and functions entirely differently than in HotA. You cannot make it a dwelling because then it would be flaggable and also grow the amount of creatures.

But you seem to know how to do it, so go ahead.

PR #388, opened 2026-01-19
Fixes and better Heat Stroke implementation

Karyoplasma:
Kinda all over the place this one, sorry, I'll go mod by mod.

mapObjects:
+ As requested in #305, I fixed the layout in the Spit creature bank.

gameBalance:
+ artifacts
    + reverted the Necromancy arti nerf (HotA 1.8.0)
    + made Badge of Courage not give mind spell immunity (arti bonuses are merged as of VCMI 1.7)
    + cleaned up the file
+ creatures
    + cross checked creature fight/AI values and updated it (HotA 1.8.0)
+ spells
    + implemented the [Forgetfulness bug](https://heroes.thelazy.net//index.php/Forgetfulness) (melee damage reduce)

cove:
+ updated creature animation to the values in monanim1 in HotA.dat
+ fixed the name of the Maelstrom in dwellings.json

factory:
+ updated creature animation to the values in monanim1 in HotA.dat

Re-implemented how Heatstroke ability functions because the old one was...:
Ability is castable 99 times in combat by the Dread/Juggernauts, only self-target.They play their multihex attack animation during casting and I dunno how to tell them not to, so I gave them the Bloodlust overlay while casting to distinguish it from the real attack. The buff they get sets their stack speed to 0 and unlocks no enemy retaliation and the multihex attack for only this combat turn. Gave them their normal attack sound back because they use their "shoot" sound when casting their buff.
Also fixed their Multihex Animation and they actually use it now when they hit multiple targets with Heatstroke.
It's as close as I could get to the HotA version and much better than what they used to have (always no enemy retaliation and always multihex attack, but without the animation).
Oh, there is 2 versions of that spell only because of the icon lol. there was a Juggernaut and a Dreadnought casting Heatstroke as, I think, an ability icon that I thought they are really cool. So I've cut out their silhouettes and placed them on a blank spell effect icon.


by003:
I have tried it before, but it is really funny that the creatures will cast a heat stroke animation before they actually carry out the attack.

Someone told me that png+json files can config more animations than def files. But I haven't tried. It might be the only way to distinct multi-attack and spell casting.

Anyway, AI is too stupid to use it.

PR #389, opened 2026-01-19
Fix for Ancient Altar Old

misiokles:
Anyone knows map name with this Ancient Altar Old object?


GeorgeK1ng:
Finding Leena II and few others.

PR #394, opened 2026-01-20
Fixups and Couatl ability refactor

Karyoplasma:
+ Refactored the (Crimson) Couatl ability spell
+ Updated the description on that ability

It was unnecessarily complicated and completely illogical: why non-magical (does not affect non damage spells), why so many bonuses that contradict each other (dispel+dispelhelpful immune into receptive makes 0 sense). I removed all of that and this is functionally identical (when https://github.com/vcmi/vcmi/issues/6762 is fixed) to what it was before. It is simply impossible to make a buff undispellable, you can even change Dispel's targetCondition to exclude Couatls, it doesn't matter, it will ignore that.

+ Added Tar Pit ambient sound (@misiokles )
+ Corrected Beholders' Sanctuary name
+ preconfigure support for the Naga Bank we (unnecessarily) overwrite (it gives gems in core as well for whatever reason)

### DIFF
--- a/hota/Mods/factory/content/config/hota/factory/creatures/crimsonCouatl.json
+++ b/hota/Mods/factory/content/config/hota/factory/creatures/crimsonCouatl.json
@@ -59,7 +59,8 @@
 			"castsymMeditation" : {
 				"type" : "SPELLCASTER",
 				"subtype" : "meditation02",
-				"val" : 0
+				"val" : 0,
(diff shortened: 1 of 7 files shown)


by003:
This ability was written from the early age without invincible. Thus all factors should be considered. Now it's pretty good for only use one type to finish it 

I still think the ability should be added as a unique label, they should be able to make the AI use this ability compulsorily in the first turn.

PR #403, opened 2026-01-22
Fix all terrain map objects

Karyoplasma:
+ made Garden of Revelation spawn on all terrains (couldn't spawn on snow, mod terrains and at least one more core terrain, probably sand)
+ made Magic Well spawn on all terrains (couldn't spawn on mod terrains)
+ added a redwood observatory template to highlands and wasteland (couldn't spawn there)

I don't think the first 2 are hota-specific fixes. Garden and Magic Well should spawn on all terrains to my knowledge, the Redwood Observatory might be a different case because lava and subterra should spawn Pillar of Fire instead which is an entirely different object, so we might have to add those manually.


Karyoplasma:
found a whole bunch of broken masks, will send new PR

PR #408, opened 2026-01-25
Lots of bug fixes

### LINKED ISSUE #296: Giselle specialty does nothing
Replication steps:
1) Make a map with your lvl20 Giselle (82k XP) and an enemy hero with 50 Spell Power.
2) Fight the hero.

Expected Result:
Enemy Spell Power is reduced to 20 
Formula: -30% * (1 + 0.05 * 20) = -30% * 2 = -60%
-> 50 * 0.4 = 20

Observation:
Enemy has 35 spell power (50 * 0.7 = 35)

Conclusion:
The Interference specialty does nothing.

I tried to fix it, but I don't really understand the result. You can add "main" to the specialty field in interference and then use `"secondary" : "interference"` and the spell power is reduced to 34, but this does not grow with level at all (lvl1 Giselle with Expert Interference also reduced to 34). If you set `"stepSize" : 100`, you simply double the effectiveness of Interference and it respects the skill level, but not the hero level. I could not get a bonus to scale with hero level at all. I assume it's because Interference uses `PERCENT_TO_ALL`, which it has to, else the effect won't work at all.

I tried to understand the updater formula, but it requires someone who is better at math than me.

Karyoplasma:
I have tested all of these changes and they function as expected.

**major fixes:**
+ Made Cannon actually able to target siege walls
    + I don't know how to change the arced shot animation of catapult-type spells, but it works at least
+ Fixed Eovacius clone being spammable in combat (now can only be cast once per battle)
    + Spell cost back to normal
    + Added spell scroll icon for spell book right click
+ Fixed Giselle's specialty (fixes #296 )
    + Also default-banned Resistance when interference is enabled
+ Reverted Garden of Revelation and Magic Well spawning on all terrains. I was mistaken.
    + Added Garden of Revelation and Magic Well to highlands terrain
    + Added Magic Well to Wasteland terrain (Garden cannot spawn there)
+ Fixed missing basic and advanced values in Ballistics skill and add Cyclopes (and Kings) missing values

**small fixes:**
+ Added Runes to the banned skills in Witch Huts and Universities
+ Fixed Sea Witch/Sorceresses spell after attack (casts only after shooting)
+ Fixed allowed terrains of the Skeleton Transformer map object
+ Fixed the cost of some artifacts
+ Switched the bonus of Cape of Silence to the newer `"BLOCK_MAGIC_BELOW"`

**animation fixes:**
+ Added Terrain Limiter for Sandworms/Olgoi-Khorkhoi teleport ability
+ Fixed moveStart and moveEnd animation of Sandworms/Olgoi-Khorkhoi when they cannot burrow
+ Changed Juggernaut and Dreadnought casting animation to make it less confusing 
    + they play their mouseover animation now where they pump their arms and release smoke

**probably temporary changes:**
+ Changed Grave to also make the digging hero unable to move the next turn. Added disclaimer about that to message.


Karyoplasma:
discovered bug

PR #417, opened 2026-01-30
Templates

Karyoplasma:
Added a bunch of templates, including a Jebus Cross with HotA settings (3 maxperzone hermitshack, mineralspring and tavern, enabled Ancient Lamps in starting zones)

Missing:
+ Firewalk
+ Clash of Dragons
+ Anarchy pack


GeorgeK1ng:
Oh nice! Will check it later today when Im back on PC!

Karyoplasma:
Some templates (Black'n'Blue, Nine-day Wonder as examples) are a bit different as in HotA. We cannot ban/spawn specific rewards in Pandora Boxes or Prisons and we would also need a functionality for creature-tier dependent dwelling spawns. Most of them are good tho.

PR #426, opened 2026-02-08
Animation fixes

Karyoplasma:
+ fixed shooting animations (attackFrameClimax and offsets) of pretty much every ranged unit introduced in HotA
+ made Halfling Grenadiers use their special projectile animation
+ fixed walking and casting animation speeds of neutral creatures
+ limited bonus luck of Leprechauns to positive values only, removed the fixme comment
+ implemented HotA animation changes to vanilla creatures
+ moved pirate projectile def into hota/creatures/battle/projectiles
+ made Gunslingers/Bounty Hunters use the pirate projectile instead of the halfling rock


IvanSavenko:
> The fix would be to copy over the Pirate projectile from cove and tell them to use that, or to hard-depend on cove and link them to the already existing projectile there. Dunno if I should change that.

Or you can put that asset into hota mod itself (which is currently basically empty). So it will be always available for all hota submods

Karyoplasma:
> Or you can put that asset into hota mod itself (which is currently basically empty). So it will be always available for all hota submods

That is the best solution, I'll do that.

PR #429, opened 2026-02-09
Fixed handling of Runes skill mechanics

IvanSavenko:
Should make Runes mostly functional. Requires https://github.com/vcmi/vcmi/pull/6850

Based on @Karyoplasma branch. Includes some other changes, not sure which are needed and which are not, so please check this PR and either merge it or simply grab changes and open better PR.

With this PR Runes skill should now correctly give bonuses and show up in UI. Runemaster is unchanged, so still broken, but fixable by copy-pasting Runes logic


Karyoplasma:
Tested, it works.

I'll merge this and fix Yeti and stuff in another PR.

PR #436, opened 2026-02-10
Submod to ban map-editor-only alternative sprites

Karyoplasma:
This PR introduces a new submod gameBalance.objects.disabledDecorations that disables some objects from mapdecorations from spawning:
+ double-sized wagon
+ mini-prison and hero camp
+ subterranean gate alternative

Others can just be added here and the user can choose whether they want the additional decorations or not.

Optimally, this would also disable biomes, but any attempt to do so was just crashing VCMI, so instead I disabled the most offensive biomes in mapdecorations (destroyed camps/houses/fountains).

### DIFF
--- a/hota/Mods/gameBalance/mods/objects/mods/disabledDecorations/content/config/hotaGameBalance/disabledDecorations.json
+++ b/hota/Mods/gameBalance/mods/objects/mods/disabledDecorations/content/config/hotaGameBalance/disabledDecorations.json
@@ -1,0 +1,42 @@
+{
+	"core:prison" : {
+		"types" : {
+			"prison" : {
+				"templates" : {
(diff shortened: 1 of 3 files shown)


GeorgeK1ng:
So this settings in biomes leads to crash instead purge it? **"ruinedObjects" : null**

Karyoplasma:
> So this settings in biomes leads to crash instead purge it? **"ruinedObjects" : null**

Exactly what i tried, crashes. I also tried just overwriting the templates, which would only throw a warning that could be ignored, but that crashes as well. Uncommenting/deleting in mapdecorations is the only thing that disables them without crashing.

GeorgeK1ng:
Let's do bug report in VCMI repo. Merging for now.

PR #456, opened 2026-02-13
Updated all artifact descriptions

Karyoplasma:
Updated all artifact descriptions. Made a new file in heroes3DataPatch for the artis that only had a description change since this is not a balance concern.

Also fixed a typo that lead to Monks and Zealots using the wrong projectile hit SFX in changedShootingVisuals.

### DIFF
--- a/mods/artifacts/Content/config/hotaArtifacts/capeOfSilence.json
+++ b/mods/artifacts/Content/config/hotaArtifacts/capeOfSilence.json
@@ -20,7 +20,7 @@
 		},
 		"text" : {
 			"name" : "Cape of Silence",
-			"description" : "{Cape of Silence}\r\n\r\nWhile wearing this cape, neither you nor your opponent will be able to cast level 1-2 spells during combat.",
+			"description" : "{Cape of Silence}\n\nWhile wearing this cape, neither you nor your opponent will be able to cast level 1-2 spells during combat.",
 			"event" : "As you pass the graveyard, you notice an old man mourning over his wife. As you approach him, the old man falls dead. After burying him, you decide to take his cape."
 		}
 	}
--- a/mods/artifacts/Content/config/hotaArtifacts/charmOfEclipse.json
+++ b/mods/artifacts/Content/config/hotaArtifacts/charmOfEclipse.json
@@ -26,7 +26,7 @@
 		},
 		"text" : {
 			"name" : "Charm of Eclipse",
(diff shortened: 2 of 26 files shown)


GeorgeK1ng:
This is exactly change which is going to break Chinese, Czech and probably German user experience right now for 1.7.2 release...

Lets keep this for next week. Will create then another branch for these changes, to not put them directly into Launcher.

Karyoplasma:
How is this going to break anything? It will still display the current descriptions in those languages... they are then just not updated. Unless one plays on 2 languages simultanously, they won't even notice this lol

Edit: ah no, apparently changing a description overwrites the translation of that field. Damn talk about crap encapsulation. This is a design flaw tho.

GeorgeK1ng:
Changes from artifacts.json and artifactsDescriptions.json overwrite them.

Karyoplasma:
is the same true for spells?

GeorgeK1ng:
> How is this going to break anything? It will still display the current descriptions in those languages... they are then just not updated. Unless one plays on 2 languages simultanously, they won't even notice this lol
> 
> Edit: ah no, apparently changing a description overwrites the translation of that field. Damn talk about crap encapsulation. This is a design flaw tho.

Nope, thats perfect design, every touched string require same touch for translations. Perfect for multilanguage support, otherwise you will do changes and every other translation need to seek in mods...

Karyoplasma:
>Nope, thats perfect design, every touched string require same touch for translations. Perfect for multilanguage support, otherwise you will do changes and every other translation need to seek in mods...

I don't see how that is good design in any way, shape or form. It's horrible.

Anyway, I'm gonna make a PR with the projectile hit sound fix and merge it. This one can wait until after release.

IvanSavenko:
> Edit: ah no, apparently changing a description overwrites the translation of that field. Damn talk about crap encapsulation. This is a design flaw tho.

By design. If a mod (like hota in this case) modifies description (for example, because effect of skill/artifact/specialty) was changed, game will ALWAYS pick string from the mod, and will ignore string from translation mod. Strings that are modified must be translated in the same mod.

This fixes confusing cases where hota mod had modified descriptions for skills like Estates (due to skill buff) but player saw old skill description because he is using translation mod for the base game.

So such strings must be translated in the same mod.

As far as this PR goes:
- fixing names of hota artifacts is safe - translations will stay active (but may need an update, to account for changes)
- fixing names of h3 artifacts is not safe and will disable translations for these strings

PR #461, opened 2026-02-14
Tark portraits fixed

Aph-ra:
Transparent/random black pixels removed


Karyoplasma:
Thanks. They were directly from the hota png folder, so they must be wrong in HotA as well xD

Aph-ra:
No worries! There have been transparency issues in the past with other portraits (ROE Catherine for example) and other HOTA elements, so if I find any more, I'll make new pull requests :)

PR #467, opened 2026-02-14
Fix bonus of Necromancy Amplifier

Karyoplasma:
I erroneously changed the Necromancy Amplifier back to 10%, I thought 1.8.0 changed that too, but it's still 5%.


Karyoplasma:
The validation fail is a bug in the validator.
`Template animation file hota/factory/creatures/adventureMap/AVWcoat was not found!`
It is found. The game displays it correctly.

IvanSavenko:
> Template animation file hota/factory/creatures/adventureMap/AVWcoat was not found!

Yeah, already have fix locally, but haven't pushed it yet. Bug in vcmi, not in hota

PR #481, opened 2026-02-16
Moving the polished strings into translations

Karyoplasma:
This is a work-in-progress.

TODO:
[x] artifacts
[x] objects
[x] spells

Additional notes:
artifacts:
I have also commented out the outdated strings the gameBalance.artifacts translations. I did NOT check whether these are added in the global hota translation. They should be there, so if they aren't please add them there with the original 5/10/15% values.
Equestrians and Boots of Speed had value changes, so that should be reflected as well (or use the global translation which probably does not add the movement value).

objects:
Requires core update because some of the objects there point to the wrong entry in xtrainfo. Also hillFort cannot be localized yet.

spells:
I added versions for spells that are changed in gameBalance to the strings in the base game as well. Like for example unnerfed Summon Earth Elemental that does not allow to summon other Ellies. More complete this way.


Karyoplasma:
@GeorgeK1ng

This would be pretty much the final version. Should not conflict with translations, I don't add any descriptions.

However, other languages might want to add something like this as well.

Karyoplasma:
i fucked up my push because i'm stupid, redoing

PR #483, opened 2026-02-17
Fixed crash on loading Seafaring Academy on maps with banned Navigation

### LINKED ISSUE #6909: Forged in Fire campaign can't be played
Start HotA, Forged in Fire campaign, use cheat for first 2 campaigns. Try to begin next scenario.

Image

[VCMI_Client_log.txt](https://github.com/user-attachments/files/25351810/VCMI_Client_log.txt)

IvanSavenko:
NOTE: not bumping version in case if there other planned changes, like rest of currently open PR's

While these two configs appear to have same logic, there is one important distinction:
```json
    "anyOf" : [
        "navigation"
    ]
```
This means: out of skills available on map, select one of the following skills: navigation.
```js
    "navigation"
```
While this means, "select navigation"

Normally, these are identical. But if map has Navigation banned, first code will fail since there are no skill available on map that satisfy this logic.

- Fixes https://github.com/vcmi/vcmi/issues/6909

### DIFF
--- a/mods/mapObjects/content/config/hotaMapObjects/seafaringAcademy.json
+++ b/mods/mapObjects/content/config/hotaMapObjects/seafaringAcademy.json
@@ -9,11 +9,7 @@
 					"resource-skill"
 				],
 				"offer" : [
-					{
-						"anyOf" : [
-							"navigation"
-						]
-					},
+					"navigation",
 					{
 						"noneOf" : [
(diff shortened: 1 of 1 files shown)


Karyoplasma:
I remember changing that, yes.

The reason for that change was a schema warning when I didn't use anyOf. Seems to be fixed.

PR #492, opened 2026-02-23
Cleaning up duplicates and unused files

Karyoplasma:
There are a lot of duplicate assets throughout the mod that could be cleaned up.

This is a draft for now because there are a lot more, I only cleaned up the safe stuff for now. 

Status: Virtually all adventure map creature sprites are duplicates in heroes3DataPatch and newGraphics. Also, as Ivan already said the adventure map town sprites are duplicates between those mods as well and pointed out that heroes3DataPatch doesn't even use them properly (the configs for this are in newGraphics).

What should we do? Which mod should do what?

@GeorgeK1ng there are also some translated voices that are duplicates:

  > maps\Mods\polishVoice\content\sounds\FACT11.ogg
  > maps\Mods\polishVoice\content\sounds\FACT12.ogg

Same size, same SHA-256


GeorgeK1ng:
Yes, thats correct for campaigns mapping in map support / campaigns override..

Karyoplasma:
> Yes, thats correct for campaigns mapping in map support / campaigns override..

Okay, we'll leave those. 

But the files I have deleted in this PR were just dummy duplicate sound files for the Bulwark creatures (Yetis and Runemasters use the same sounds for example) and some weren't even set to be used in config and some duplicate ability icons, so this can be reviewed.

PR #504, opened 2026-03-02
Resolved z-index issues in towns

Karyoplasma:
This PR resolves the z-indices of overlapping buildings or changing z-indices in building upgrades throughout towns modified or introduced by HotA. Doing so fixed the display of the Castle Gate (now properly blends into background) and the Vault of Ashes (resolved the clipping of the decorative waterfall). I did not notice any visual changes in the other towns, so maybe we got lucky there.

I did not move any buildings (don't think there is anything wrong there), only adjusted z-indices.

@by003 you have a keen eye for details, can you check if I overlooked something?

Inferno: 

before:
infernobefore

after: 

infernoafter

Conflux:

before:
confluxbefore

after:

confluxafter

Cove:

before:
covebefore

after: 
coveafter

Factory:

before:
factorybefore

after: 
factoryafter

Bulwark:

before:
bulwarkbefore

after: 

bulwarkafter

### DIFF
--- a/mods/bulwark/content/config/hota/bulwark/town/structures.json
+++ b/mods/bulwark/content/config/hota/bulwark/town/structures.json
@@ -92,7 +92,7 @@
 					"campaignBonus" : "hota/bulwark/town/icons/campaignBonuses/marketplace",
 					"x" : 397,
 					"y" : 249,
-					"z" : 0
+					"z" : -1
 				},
 				"resourceSilo" : {
 					"animation" : "hota/bulwark/town/buildings/TBBuSilo",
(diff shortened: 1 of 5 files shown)


by003:
I haven't found any other visual issues.

I can't know the exact value of the z-coordinate either. When I first started doing Bulwark screen, I just tested to see if there was occlusion. I would increase the z-value by 1-2 if occlusion had been found.

by003:
Actually, no matter a building's z-value is 5 or 99, if the overall look is harmonious, that's fine.

PR #520, opened 2026-03-14
Update Russian translation

Karyoplasma:
This PR fixes the issue of some objects having an English message (Scholar and Arena for example) even without activating the gamebalance.objects submod that changes them.

I have also added the translations for the missing gamebalance.skills submodules. It uses the skill names of the core game and I snipped the parts when HotA referred to them with their new names.

Additionally, I added all strings that should be translated to the parent mod's translation file. In the cases where it was sensible, I added the translation from the HotA.dat, some are still untranslated, but they are now easier to change since they already exist in the correct place.

I only used AI for things like the Mana Vortex message. In HotA, it refers to the hero by name using %s. This is not possible in VCMI, so I asked ChatGPT to replace the %s with "your hero" and keep the rest as is.


Fabrice-TIERCELIN:
@Karyoplasma, when I install _Horn of the Abyss_ & the Russian translation, I hear the English dubbing instead, whatever the order of the installation. Is it working for you?

Karyoplasma:
> @Karyoplasma, when I install _Horn of the Abyss_ & the Russian translation, I hear the English dubbing instead, whatever the order of the installation. Is it working for you?

gog.com russian installer -> HotA (choose russian localization)

only 2 of the campaigns have voice. All-In and Forged in Fire, rest is silent (at least intro video).

GeorgeK1ng:
VCMI have generated missing Russian voice. You should hear it correctly, when you have VCMI in Russian and installed and enabled Russian language mod.

Same approach goes for Czech / Polish / German.

English generated audio is used always as default.

Karyoplasma:
> VCMI have generated missing Russian voice. You should hear it correctly, when you have VCMI in Russian and installed and enabled Russian language mod.

Should I replace the generated ones with the ones from the original localization should they exist? Are we allowed to do that even?

GeorgeK1ng:
If they add them, then replace, currently we simply filled audio gaps. Currently is Russian mixed 2 original ones, 3 generated.

Karyoplasma:
Then it's already up to date. Only 2 HotA campaigns are voiced in original.

Fabrice-TIERCELIN:
> @Karyoplasma, when I install _Horn of the Abyss_ & the Russian translation, I hear the English dubbing instead, whatever the order of the installation. Is it working for you?

@Karyoplasma, @GeorgeK1ng & @heroesiiifan, I have found the bug. `russianVoice` submod has a [dependency on `mega-pack-rus`](https://github.com/vcmi-mods/horn-of-the-abyss/blob/vcmi-1.7/mods/maps/Mods/russianVoice/mod.json#L9) but when I install `mega-pack-rus` mod, it creates a folder `russian-translation`. If I manually change the dependency to `russian-translation`, I hear the Russian dubbing.

@misiokles told me not to rename `mega-pack-rus` to `russian-translation` but now that the problem is already here, I think that the best option is to rename [`https://github.com/vcmi-mods/mega-pack-rus`](https://github.com/vcmi-mods/mega-pack-rus) to `https://github.com/vcmi-mods/russian-translation` and change all the dependencies to `russian-translation` (doing so, I advise to remove the useless folder `russian-translation` inside it too).

GeorgeK1ng:
Mod rename is not good idea right now. Will check option for some kind of "obsolete" attribute for mod.json, because when we rename it will be installed twice, so mechanism to remove old one is needed first. 
I just tried to install Russian via Launcher, and is true, it's russian-translation. Just depency is needed to correct in voice and main menu configs.

IvanSavenko:
> Mod rename is not good idea right now. Will check option for some kind of "obsolete" attribute for mod.json, because when we rename it will be installed twice, so mechanism to remove old one is needed first.

What we need is some sort of migration support in mod repository itself.
Currently all mods are listed in `availableMods` field in repo 1.7 json
We need a separate field named `migratedMods` or something, that should contain mapping of old mod -> new mod that must be installed instead

misiokles:
Or we can just change "depends" from "mega-pack-rus" to "russian-translation" where needed.

GeorgeK1ng:
Fixed in: https://github.com/vcmi-mods/horn-of-the-abyss/pull/596

PR #522, opened 2026-03-14
English text fixes

Karyoplasma:
Some fixes for English texts, mostly done in textPolishing, although this might want to be moved into heroes3DataPatch as they are actual fixes, not just wording adjustments.

Example:
Glyphs of Fear incorrectly states it boost defense instead of attack, Conflux town calls every upgraded building a wrong name, etc. HotA changes these strings, so the information is presented correctly. Would it be better to move these into heroes3DataPatch and with a localized string, not as a "translation"?

Opinions? @misiokles @by003


GeorgeK1ng:
What's example of Conflux buildings names changes? Some fixes can be applied to language pack directly - but it don't mean, english players have it installed.

Karyoplasma:
>What's example of Conflux buildings names changes? Some fixes can be applied to language pack directly - but it don't mean, english players have it installed.

"building.core.conflux.dwellingUpLvl4.description" : "The Temple of Energy allows you to recruit Energy Elementals." 
->
"building.core.conflux.dwellingUpLvl4.description" : "The Upg. Altar of Fire allows you to recruit Energy Elementals."

The building is named "Upg. Altar of Fire" in SoD as well. The building names are all what you'd expect, just the building descriptions call them differently for some reason.

We could also just load the file with the text bugfixes in heroes3DataPatch, requiring English Launcher setting. That should work as well, although that might be overwritten by the English Translation Mod, so maybe a soft-depends on it.

GeorgeK1ng:
That's exactly something what we need to apply here too as it's typo / logic fix for original translation. https://github.com/vcmi-mods/h3-for-vcmi-englisation/blob/vcmi-1.7/H3forVCMIenglisation/content/config/h3forvcmienglisation/game.json

Karyoplasma:
> That's exactly something what we need to apply here too as it's typo / logic fix for original translation. https://github.com/vcmi-mods/h3-for-vcmi-englisation/blob/vcmi-1.7/H3forVCMIenglisation/content/config/h3forvcmienglisation/game.json

It's not too many. Almost verything in the gameTextChanges.json is like that. Some are HotA-only (like Thatched Hut -> Halfling Hut), but most are either typo-fixes or fixing some silly stuff like above.

EDIT: I've sent a [PR](https://github.com/vcmi-mods/h3-for-vcmi-englisation/pull/20) with the most non-controversial fixes. Literally just typos and logic errors.

PR #525, opened 2026-03-15
Latvian translation

Fabrice-TIERCELIN:
Hi @misiokles,

This PR translates the mod in Latvian.


misiokles:
@Karyoplasma 
Is it safe to merge after latest re-organisation of submods?

GeorgeK1ng:
Yes, this is only base translation. Submods are not included.

Karyoplasma:
> @Karyoplasma Is it safe to merge after latest re-organisation of submods?

Well, displayed information will be wrong if gamebalance is not activated, but it's AI slop anyway so I don't really care. (Necro artis are wrong as well).

It won't break existing translations tho if that's what you're asking.

PR #531, opened 2026-03-18
Fix accidental removal

Karyoplasma:
@Fabrice-TIERCELIN 
French translation has 2 left-over Russian strings in gamebalance.creatures
Also, please make sure that the text polishing mod description in the languages say that it's English-only, not that the user activates it and expects a change.


Fabrice-TIERCELIN:
I will fix it

PR #541, opened 2026-03-22
Fix stuff in German translation

Karyoplasma:
+ didn't realize "Ödland" refers to rough terrain, referring to wastelands as "Einöde" now
+ replace cartographer messages with original messages, only adjusting gold value
+ replace shrine lvl 4 text with something more in line with the original shrine's event texts
+ added Spirit of Oppression and Hourglass override for stating that they negate positive effects only


Fabrice-TIERCELIN:
@Karyoplasma, I'm not a German speaker but I suspect that [_"Alles drauf"_](https://github.com/vcmi-mods/horn-of-the-abyss/blob/vcmi-1.7/content/config/translation/hota/german/campaigns.json#L77) is a wrong translation. In English, _All In_ both means:
- Having a room booking with all the services included
- Use everything we have in a fight, so we win or we lose everything

In Russian, _Ва-банк_ is not clear: GLM AI says it means the second meaning. Google Translate says it means the first meaning. Translation sites say it clearly means the second meaning.

It's a French alternative translation that warns me that the French VCMI translation was wrong.

Karyoplasma:
"Alles drauf" would be an all-in attack order, but it does sound pretty colloquial. "All In" as in "all inclusive" would be more like "Alles drin" or something like that, no possibility of misunderstanding there.

I assume it should be an attack order as that would make more sense.

PR #546, opened 2026-03-24
Fix Navigation specialists

Karyoplasma:
Apparently, the text of Navigation specialty is wrong and it simply doesn't give a 5% bonus to Navigation skill. Rather, it grants 75 sea movement per level, regardless of skill level.

I tested in both SoD and HotA and confirmed that HotA didn't fix this (even tho they changed the specialty text to a different variation of 5% lol)

Related PR: https://github.com/vcmi/vcmi/pull/7139


Fabrice-TIERCELIN:
Hi @Karyoplasma,

Which translation is accurate (or none)?

_Я снова вспоминала дом. Те мирные, сонные времена, когда даже редкие драки на деревенских гуляниях всегда заканчивались мировой чашей…_

_I thought back to home again. Those peaceful, sleepy times, when even the rare fights at village festivals always ended in a **bloodbath**…_

_I thought back to home again. Those peaceful, sleepy times, when even the rare fights at village festivals always ended in a **chalice of peace**…_

Karyoplasma:
I don't speak Russian, but sentence structure alone suggests that they ended peacefully, so "bloodbath" does not fit at all. 

"chalice of peace" makes no sense tho, what they are doing in my head is to reconcile after a tussle, sitting down for a beer or so, so maybe a "drink to peace"?

Honestly reminds me of Asterix lol

PR #575, opened 2026-04-10
Highscore creatures submod

### LINKED ISSUE #497: Create High Score creatures
In HotA, there are more high score creatures can be choosed.

But I cannot find any website to show them.

Karyoplasma:
Fixes #497

I used your list from that thread @by003 

Took some time because I'm still nerding Balatro lmao

Also, should this mod be default-disabled since it depends on all factions and neutralcreatures? Or is that fine?


by003:
Well done!

Actually it can be disabled in normal situation, but I don't think it will have much impact. This mod only relates to the three new factions in HotA and the newly added neutral units.

Moreover, I don't think anyone would install only half of HotA... like disabling a certain faction or creature. That way, even HotA's maps and campaigns wouldn't be usable.

GeorgeK1ng:
This submod should be outside data patch. Data patch is for OH3 files.

Karyoplasma:
> This submod should be outside data patch. Data patch is for OH3 files.

But where to put it? It's not balancing, it's not really new graphics. Standalone submod, I want to avoid, we already have so many, it is becoming very confusing.

GeorgeK1ng:
I would place it on same level as Bulwark / Data patch. And instead to make it visible change it to Compatibility and it will be automatically enabled when all dependat mods are active. So it will be invisible and automaticaly activated when all 3 towns are enabled.

PR #581, opened 2026-04-12
Fix elementals specialty overlap

GeorgeK1ng:


Karyoplasma:
Should be fixed in the engine, not in configuration. Reverts an intentional change as a band-aid fix.

GeorgeK1ng:
Not in engine, every language have it's own rules for string cropping. Must be done by strings editing.

Karyoplasma:
VCMI ignores line breaks in specialty names, tho, I tried that. So it should be fixed in the engine. Also the behavior in HotA is different in the hero pick screen where there is plenty space from the missing morale/luck info boxes. It fits there anyway, so nothing should be done. Only in the hero info screen ingame is where the length really matters. This PR probably also misses Solmyr, although I haven't checked that, but "Chain Lightning" was another one of those abbreviated specialties.

GeorgeK1ng:
I think I have found reason, why it's ignored \n in hero specialty. Will test it and open PR.

GeorgeK1ng:
Well, this needs more changes than expected. Adding multiline support is easy, but it won't fit easily there without breaking other UI elements. And in HotA (HD Mod) it probably fits there due different font? Or did they rebuild Hero Window Background image?

image

IvanSavenko:
Hota uses same background image, but they use smaller, sans-serif font (poor fit for game like h3 imo). So yeah, longer texts may fit in hota

PR #584, opened 2026-04-14
Cove & Factory in French & Latvian

Fabrice-TIERCELIN:
Hi @Karyoplasma,

This PR translates _Cove_ & _Factory_ into French & Latvian.


Karyoplasma:
I kinda had the idea in the back of my mind to shift the other translations to this structure as well (translating mod by mod instead of having one huge file in the parent mod).

Seems cleaner and when something is missing, you don't need to figure out where it should be added, you can just add it in the mod that adds the content.

Thoughts? @GeorgeK1ng

GeorgeK1ng:
No, just add it only to mods which require it, best to keep it on one place when possible.

Run **/translate** command in game and you get correct structure. We have also **/translate** missing commands to track missing strings. It's easy.

Can't Imagine pain when need to translate 20+ submods. It's already anoying to add translation name and description into each submods mod.json. Also Its easy to check translation consistency during reading lines in one file instead of switching in gazilions of files.

Karyoplasma:
>Run /translate command in game and you get correct structure. We have also /translate missing commands to track missing strings.

Problem with this is that it doesn't respect changing strings. Example, all heroes are set to be translated in hota.json, however it needs to be translated twice because of gamebalance.heroes changing it. It never suggests to translate gamebalance.heroes, so could lead to wrong values on display.

GeorgeK1ng:
It will correctly gives you 2 files. HotA.json a hota_gamebalence.json or something like that when you have enabled game balance during /translate commands. This commands exports strings of enabled submods. And yes, Its needed to translate it twice, but Heroes game balance have only strings which were modded by that submod.

Fabrice-TIERCELIN:
Honestly, I have put those files here thinking to myself : "It can be moved to any better place later by anybody" ^^

Issue #591, opened 2026-04-16
Writing biographies for some special heroes


by003:
There are six factory heroes are used in campaign 'Forged in Fire', they don't have any biographies and I just used the description from Wiki, like: 
%s is a defender of Burton once Kastore took control.

Now I have written biographies here.

**Jangaard**
Jangaard, inherently cruel, was once feared throughout Raven Shore's pirate crew. He now appears to have joined Kastore's guard, serving his ruthless master. On the battlefield, Jangaard always delivers a surprise attack.

**Stina**
Stina, who has always lived in Burton, is willing to cooperate with anyone for survival. This girl is incredibly brave; it is said she dared to face over a dozen raiders alone in the desert and repel them.

**Valquest**
Valquest, who traverses the desert, possesses exceptional medical skills. He has joined Kastore's ranks and is teaching his skills to others. Although he appears cruel, he has a compassionate heart for small animals.

**Winzells**
Winzells, who always considers himself a gentleman, pays particular attention to his appearance, constantly sourcing the finest fabrics to make his clothes. However, when he appears on the battlefield, he is anything but gentlemanly.

**Elderian**
Elderian, living in Enroth, seeks to research forbidden magic. He has visited many scholars but cannot find what he truly seeks. It is said that he can complete his research in Burton. He wouldn't care even if it meant joining the ranks of evil.

**Umender**
The elderly Umender had spent much time studying mechanical operation; he was an expert in the field. It was precisely because of this that, after arriving in Burton by chance, he developed many strange mechanical devices for King Kastore. But he was not satisfied with his current achievements.

@Karyoplasma @GeorgeK1ng 
You can also try to localize these things into your own language.

Karyoplasma:
English + German are done, rest of the languages still have the placeholder biographies.

alekgosk:
I added the polish translations in #597

Karyoplasma:
Since we use Weblate now and the biography changes are in, this can be closed

PR #593, opened 2026-04-16
Missing biographies for campaign-only heroes (+German translation)

Karyoplasma:
Addresses #591


alekgosk:
@Karyoplasma I added polish campaign texts in #597

PR #600, opened 2026-04-22
Delete iViewCr2.def + Fix Naga Bank message

GeorgeK1ng:
This was created by me to fix visual glicht, moved to VCMI Extras to deliver it accross all mods. To be merged after 1.7.4 release


Karyoplasma:
I think we don't need to overwrite the Naga Bank at all actually. core implements the HD mod reward change from sulfur to gems already, so it should just be removed from the config, I guess?

GeorgeK1ng:
Point of Naga tweak here was missing reward message. Core config should be then changed if it doesn't match OH3 config.

PR #607, opened 2026-05-01
Japanese cleanup

Karyoplasma:
@heroesiiifan

Some hero strings are missing in parent mod game.json translation file. The reason for this is that these strings are overwritten in gamebalance.heroes and if you start the game with that mod activated it will generate the translation file there, but that means it the entries will be missing if the user does not enable gamebalance.heroes. Basically, all strings that are translated in gamebalance.heroes need to also be present in the parent mod translation, but with the vanilla values (e.g. 3% spell damage instead of 10).

There also was a duplicated entry of Churchyard, one with the now invalid mapobjects.highlandsterrain... identifier: I fixed that issue, so mapobjects.highlandsterrain.creatureBanks.churchyard doesn't exist anymore, it's now always part of mapobjects. Since the translation of these were different (even tho the strings are the same), I kept both and commented one out randomly. Choose the better translation, I don't speak Japanese.

But generally, the translation was structurally clean. I just moved the core.mapobject strings into gamebalance.objects to prevent wrong information being presented when that mod is deactivated.


heroesiiifan:
Thanks!

For content itself we need to ask @Carson-N as native japanese speaker which is better.

ghost:
> For content itself we need to ask @Carson-N as native japanese speaker which is better.

・Churchyard：
Simply remove the commented-out section.

・gamebalance.heroes：
The game.json file was generated using the "/translate" command because I couldn't find the English text, and I did confirm that there are indeed several missing lines compared to files in other languages. I don't have time today, so I'll try to fix it tomorrow.

Karyoplasma:
> The game.json file was generated using the "/translate" command because I couldn't find the English text, and I did confirm that there are indeed several missing lines compared to files in other languages. I don't have time today, so I'll try to fix it tomorrow.

To be clear: that file is correct and should be there. The entries in that file should just also be present in game.json in the parent mod because else, a user that chooses to not activate the balance module will not have any translation for them. Without balance mod, all hero specialties scale as they would in vanilla, so that only affects creature specialists, spell specialists and a handful of others and only heroes that were introduced in HotA. Vanilla heroes don't need to be translated in the parent mod as they are already presumed to be translated in the Japanese translation of the core game.

PR #635, opened 2026-06-03
add polish translation file to spell balance submod

misiokles:
Adds polish empty translation file to spell balance submod to show string in weblate.


Karyoplasma:
Just as a heads up: it won't show up in Weblate this way as gamebalance.spells is not part of the Weblate project. If you want to translate it, you can copy the German translation file (has all changed strings I think). Or wait for a minute and I prepare an english.json where you can base it off of.

gamebalance.spells doesn't actually invalidate any strings, it just adds a custom version of the description for some spells to reflect the changes (e.g. removing the part where the summon spells state you can only summon one type of elemental).

Edit: it's up now.

misiokles:
@Karyoplasma I've added (manually) polish translation for spell balance submod. Guess it's safe to merge it now.

PR #640, opened 2026-06-06
Eagle Eye update

Karyoplasma:
This PR adds the chance to learn a spell pre-battle to Eagle Eye.

I changed the description a bit to strip out the parts that are not implemented yet. Originally, the hero should learn a spell during battle immediately when used instead of after and should negate 1, 2 or 3 levels of opponent's Eagle Eye just like Tactics. Also the "I can check enemy spell book thing" is missing.

### DIFF
--- a/mods/gameBalance/mods/skills/mods/eagleEye/content/config/hotaGameBalance/skills.json
+++ b/mods/gameBalance/mods/skills/mods/eagleEye/content/config/hotaGameBalance/skills.json
@@ -1,47 +1,76 @@
 {
 	"core:eagleEye" : {
-		//"base" : {
-		//	"effects" : {
-		//		"main" : {
-		//			"type" : "LEARN_BATTLE_SPELL_CHANCE",
-		//			"valueType" : "BASE_NUMBER"
-		//		},
-		//		"val2" : {
-		//			"type" : "LEARN_BATTLE_SPELL_LEVEL_LIMIT",
-		//			"valueType" : "BASE_NUMBER"
-		//		}
-		//	}
-		//},
-		"basic" : {
-			"description" : "{Basic Eagle Eye}\n\nBasic Eagle Eye gives your hero a 40% chance to learn any 1st or 2nd level spell used in combat.",
+		"specialty" : [
+			"preBattleChance"
+		],
+		"base" : {
 			"effects" : {
-				"main" : {
-					"val" : 10
+			//	"main" : {
+			//		"type" : "LEARN_BATTLE_SPELL_CHANCE",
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
This should be optional part of Eagle Eye skill balance, not force all users with this settings.

Karyoplasma:
Very much against that, honestly. If they don't like it, they can just disable that mod anyway. The mod only changes Eagle Eye. Making another submod to change Eagle Eye, but in a different way is nonsense.

The mod didn't actually do anything besides NERFING Eagle Eye from vanilla (40%/50%/60% after battle) to 10%/20%/30% and lie to the user with a wrong description.

GeorgeK1ng:
No, we are not going to force every user with this change. Eagle Eye skill change will just receive it's own submod with this change enabled by default. It's same as we have various bonus icons submods, player can simply toogle unwanted things. I have not added engine support for prebattle logic to force Eagle Eye skill change to every VCMI HotA player.

Karyoplasma:
It is ALREADY a submod of gamebalance.skill. What are you talking about? If some dumbfuck elitist doesn't want that, he should disable Eagle Eye submod and just play fucking Vanilla, I really don't give a shit.

GeorgeK1ng:
Put submod into Eagle Eye.

Karyoplasma:
No.

IvanSavenko:
> Put submod into Eagle Eye.

Why? Hota already has dozens of submods, making not trivial to configure. And H3 Eagle Eye would be basically trash even with 100% chance.

Besides - I think no hota version even had bumped EE chance with H3 mechanics - those were changed together, so  keeping current version does not makes sense even as option to give players access to specific hota version.

GeorgeK1ng:
Eagle eye was nerfed, now it becomes overkill. And also this feature is no 1:1 to HotA one. If we have 100% same one, then fine. But this pre battle is not same logic shared with HotA.

Karyoplasma:
HotA never, ever nerfed Eagle Eye, stop hallucinating. It was vanilla until they reworked it in 1.7.2. What the submod did was a non-HotA nerf of Eagle Eye for reasons unbeknownst to mankind and lying to the user about the percentages.

Rest assured that even the more OP H3HotA version of the skill is still unusable trash because it's fundamentally terrible.

GeorgeK1ng:
Well, when there is no double change to Eagle eye, then put it there. I was thinking there was nerf to learn chance after battle and then added extented logic for prebattle.

Karyoplasma:
When they added pre-battle, they just replaced post-battle chance with a guarantee to learn when a spell is cast during combat.

PR #647, opened 2026-07-01
Fix bulwark hero loading

Karyoplasma:
Bulwark hero IDs are incorrect across the board because Boyd takes up Dhuin's ID which shifts everyone down. The result is that Bulwark heroes load incorrectly when placed in a HotA map. I adjusted that with this PR by kicking Boyd out for the moment (really have no idea what to do with him here) and correcting the IDs for the Bulwark heroes. Which leads to a secondary problem: the internal ordering we use is wrong. It's correct up to 14haugir. Our ordering:
14haugir
15akka
16vehr
17allora

correct ordering:
14akka
15vehr
16allora
17haugir

Is that something that needs to be fixed or are we okay with the wrong internal order as long as they load correctly? It would be a fairly big change with also having to migrate the references in translations for those 4 heroes. Additionally I don't know what to do about the Boyd situation. So that's why this PR is merely a draft.


misiokles:
Translations is not a problem, we have weblate to fix that changes...

Karyoplasma:
Fixed the hero ordering and portrait IDs. Portrait IDs seem to be correct outside of the 4 misordered heroes in Bulwark from what I can determine.

Not sure if the translation workflow will pick up that the hero objects simply got renamed or if the strings in the various languages have to be copy-pasted.

Boyd doesn't have an entry in the HotA.dat, so he is probably a portrait-only hero like some others and doesn't need an ID in map support other than his portrait (which has the correct ID, I tested).

GeorgeK1ng (approved):

PR #649, opened 2026-07-04
update Chinese translation

kdmcser:


GeorgeK1ng:
This translation update must be done using Weblate too. See mods době in Weblate here - HotA, WoG, ToW, Ruins, Pavilion etc ... https://github.com/vcmi/vcmi/blob/develop/docs/translators/Translations.md#translation-of-vcmi-mods

PR #654, opened 2026-07-07
Rewrite Clone Specialty into Lua, semi-fix dune battlefield

### LINKED ISSUE #7370: cloned unit effect icon is missing
**Describe the bug**
I cloned Pirates. Viewing unit info (both short and full) shows "3" with no icon, same when viewing original and clones.

**Game logs**
-

**To Reproduce**

1. clone a unit
2. view unit info of original or of a clone

**Expected behavior**
Clone effect has an icon

**Actual behavior**
Clone effect has no icon

**Did it work earlier?**
dunno

**Screenshots**

Image

**Version**

- OS: iOS
- Version: 1.7.4

Karyoplasma:
This PR is mostly just me trying to figure out the Lua API.

I have rewritten Eovacius' Clone specialty to not rely on a copied spell with all its side effects (like the blank unit state icon after casting). Instead, now he has a "SPECIALTY_CLONE" bonus and the value of that bonus determines how many casts of clone produce 2 clones per battle, so you could easily modify him so that the first 3 casts produce double the clones without changing the entire spell. To do that, we patch the core:clone spellEffect routine.
There is an error logged when casting clone when the specialty is active, not sure what's wrong, seems to have no effect on gameplay tho:
`[2026-Jul-07 16:58:20.337132] ERROR [TBB worker 2] ai - Package of type struct GiveBonus is not allowed in battle evaluation`
I have tested that the core idea works and that the "counter" resets between battles. What I haven't tested is what happens when 2 clone specialists fight each other (impossible under normal gameplay), but it *should* work from a logical standpoint.

Likewise, I have made a dummy spell that casts a random amount of quicksand patches (15-20) that are hidden from both players. This is currently still slightly bugged tho since you cannot make a quicksand patch trap units from both sides of battle (crashes VCMI). Instead, they only impede the attacker which is the player 99% of the time. It's still better than the current solution imo.

Status:
Clone Specialty:
- [x] cast "charges" work as intended
- [x] resets after battle
- [x] casting doesn't add a spell buff icon
- [ ] works in specialist vs. specialist fight

Dunes Quicksand:
- [x] correct amount of patches
- [x] dunes quicksand patches are undispellable
- [x] dunes quicksand patches impede both sides


mods/terrainOverlays/content/scripts/dunesQuicksand.lua
```diff
@@ -0,0 +1,80 @@
...
+    if battle:getUnitByPos(hex, true) ~= nil then return false end
+
+    local obstacles = battle:getObstaclesOnPos(hex, false)
+    if #obstacles > 0 then
+        for _, o in ipairs(obstacles) do
+            if o:getObstacleType() ~= ENUM.ObstacleType.moat then
+                return false
+            end
+        end
+    end
```
C3pa: You can make this terser;

```suggestion
    for _, o in ipairs(obstacles or {}) do
        if o:getObstacleType() ~= ENUM.ObstacleType.moat then
            return false
        end
    end
```

Karyoplasma: Yeah, will change, thanks.

Karyoplasma:
Any objections to merging this? The specialist vs. specialist edge case is so incredibly rare that it won't affect 99.9% of the fights anyway and is a simple fix once we can add hero bonuses in battle.

PR #655, opened 2026-07-09
Sandworm ability rewrite

### LINKED ISSUE #523: Sandworm's ability
@Karyoplasma 

This is the idea I mentioned last time, in `olgoiKhorkhoi.json`, to add:
```
"castsymMeditation" :
			{
				"type" : "SPELLCASTER",
				"subtype" : "spell.summonLarva",
				"val" : 1,
				"icon" : "factory/icons/bonuses/sandvigour.png",
				"description" : "{Devour}\nCan devours corpses for additional strikes"
			},
			"combatEvent" : 
			{
					"type" : "ON_COMBAT_EVENT",
					"subtype" : "combatEventCast",
					"addInfo" : {
						"effect" : [
							{
								"action" : "bonus",
								"targetEnemy" : false,
								"bonus" : {
									"type": "ADDITIONAL_ATTACK",
									"val": 1,
									"duration": "UNTIL_OWN_ATTACK"
								}
							}
						]
					},
					"hidden": true
			},
		"castsCount" :
			{
				"type" : "CASTS",
				"val" : 50
			},
```

This eliminates the need for larva assistance, but larva will still be summoned; this might be a temporary solution.

Then, you can also delete the ability of larva.
```
"castsymMeditation" : {
				"type" : "SPELLCASTER",
				"subtype" : "sandVigour",
				"val" : 1
			},
			"castsCount" : {
				"type" : "CASTS",
				"val" : 1
			},
```
I initially thought of using ADJACENT_SPELLCASTER to implement this, which would be more in line with the HotA setting and would prevent the corpse from being eaten remotely.

However, ironically, ADJACENT_SPELLCASTER doesn't seem to be considered a SPELLCASTER, so only SPELLCASTER can activate this ability. I hope the system will fix this issue.

Karyoplasma:
This is a WIP since there are still things that don't work properly, so consider this more of a teaser for @by003 who asked several times about this.

This PR seeks to implement the ability of Olgoi-Khorkhoi to devour corpses to gain additional attacks and get rid of the old temp ability.

Progress:
- [x] Destroy corpses the unit lands on and grant +1 attack for each corpse devoured
- [ ] Retain unused extra attacks for later use
- [ ] Play cast animation when devouring at least one corpse
- [ ] Make a version for a `SPELLCASTER` bonus that moves the worm and devours corpses on target hex

Currently, if you move+attack and the worm devours a corpse, it grants an additional attack, but the bonus isn't used until the next attack command, so still pretty buggy, but a basic functionality is there.


by003:
Hopefully it can be realized as soon as possible.

by003:
In addition to this, I also hope to fix the issue where ADJACENT_SPELLCASTER is not recognized as a SPELLCASTER by the AI.
AI ​​ignores creatures with this ability.

by003:
My thinking is this: because a long-weapon attack has recently been added to the commands, we might not be able to use Lua to recreate this. Instead, we need to add a replaceable command to the VCMI system. This command is "**Devour**," and it's feasible even if it's only used for this one creature.

I found that long-weapon attacks are enabled by default, so this Devour command can also be enabled by default, but it doesn't switch with the attack command; it switches with the movement command. Let me explain in my own words:

(This is definitely not AI; I wrote all of this myself)

1. Devour is enabled by default. Under this command, units will automatically devour corpses when they walk on them, and a counter is maintained to record the number of attacks. Although I know other mods might have spells that increase the number of attacks, these are independent of this counter.

2. This counter is independent of the Devour command; it only calculates the number of attacks gained after devouring a corpse and consumes it after the attack.

3. To disable the **Devour** command, we can click to **Disable Devour** (if we can't combine the two commands into one button, we might need two buttons: `Devour` and `Disable Devour`). I studied HotA again, which combines devour with the move command, but I think the move command in VCMI can only move and not even attack.

**Devour** does not conflict with attacks, as its function is simply to devour and increase the attack count. This count is determined in real-time because the devouring check has already been performed before the attack, so it is naturally reflected in this attack, similar to the method of HotA where a double-attack is performed immediately after devouring.

Incidentally, the devouring action can call the casting action.

**Disable Devour** simply restores the default settings. Clicking it only prevents devouring from being triggered on corpse tiles and does not conflict with attacks.

@Karyoplasma How about your opinion for this?

**Devour:**
devour
**Disable Devour:**
devourOff

@GeorgeK1ng This might require some help from Codex, similar to the long-weapon directive you completed previously. We need to add directives to **Devour** and **Disable Devour**. Ignore the AI; it can always have **Devour** enabled by default.

Karyoplasma:
The problem is mainly attack/animation queuing. Walking and eating corpses works (doesn't play the cast animation tho). Moving and attacking does not work, because the unit moves, eats corpses in front of the attacker, gets extra attacks but doesn't use them in the attack.

by003:
So I think we cannot only rely on lua script. We might need to try creating a new function similar to long-weapon.

by003:
During my time studying lua, I've discovered that many things aren't so easy to do, and lua's functionality is still limited.

Karyoplasma:
The thing is: if engine changes are necessary to make that work, it's better to expand the Lua API in such a way that it makes it possible rather than introducing another bonus. In this case, the answer is probably that a bonus would have the same issue unless it goes through completely different attack-queue logic than everything else and that would not be a good thing to add because it degrades the code base.

PR #656, opened 2026-07-09
Timed Spell specialists rewrite

Karyoplasma:
This PR patches the `core:timed` routine that applies spell specialties to timed spells to the values that HotA uses. It simplifies the process of patching every hero to have a million bonuses to simulate the specialty. It is now checked and applied on cast. As a bonus, with this one, Zilare's specialty also has supporting logic (although it has no effect should the user elect to not use the gamebalance.spells module, but I checked that it doesn't cause problems when it's not activated). It also eliminates some bugs from the old way of implementing the specialties like Mirlanda's Dragonflies casting Weakness with specialty (they shouldn't but the bonus on the hero didn't care how Weakness was applied).

Since `TIMES_HERO_LEVEL_DIVIDE_STACK_LEVEL` is bugged and way stronger than it should be in 95% of the cases, I found a way to handle this type of scaling cleanly without using that updater. Check Dalton as an example.

@GeorgeK1ng The reason why I am not merging this is because this changes the application of specialties for other mods as well, so a Weakness specialist defined in some other mod would also cast it with the HotA values when gamebalance.heroes is enabled. This might be a drawback, but on the other hand, it doesn't make sense to have the same specialties with a different effect. Still wanted to discuss it first.


Karyoplasma:
I'll merge this when there are no objections tomorrow.

GeorgeK1ng:
@misiokles?

misiokles:
Of course no objections. Definitelly we can have spell specialties with diffirent effeects!

GeorgeK1ng:
But this means, HotA patch going to patch all other mods on this area.

misiokles:
Yeah, but I'm sure that 99.9% modders want their timed spells heroes to work with HotA rules.

Karyoplasma:
I have also added cure specialty that affects stronger creatures more.

If we could make subtypes for custom bonuses (or just make them hold a string value), we could patch the specialties without affecting other mods. The cure change will only affect heroes with the `"SPECIALTY_CURE"` bonus.

IvanSavenko:
@Karyoplasma what exactly is wrong with `TIMES_HERO_LEVEL_DIVIDE_STACK_LEVEL`? How exactly these specialties work in hota and how they work in VCMI? Are H3 specialties also affected by this problem?

Karyoplasma:
> @Karyoplasma what exactly is wrong with `TIMES_HERO_LEVEL_DIVIDE_STACK_LEVEL`? How exactly these specialties work in hota and how they work in VCMI? Are H3 specialties also affected by this problem?

`TIMES_HERO_LEVEL_DIVIDE_STACK_LEVEL` updates val to val * heroLevel / stackLevel but it should be val * (heroLevel / stackLevel). Due to integer division these values are only equal when stackLevel cleanly divides heroLevel, else the first formula will yield a bigger result. Yes, this affects core, HotA just exacerbates the magnitude of the error because it uses bigger values, the problem is the same.

E.g. A level 20 Adela casts Bless on an Arch Angel. Bless speciality in vanilla is 3% per heroLevel/stackLevel.
What the damage boost should be: 3 * (20/7) = 6%
What the damage boost currently is: 3 * 20 / 7 = 8%

In HotA, they've increased it from 3% to 10%, now the error is even larger:
What the damage boost should be: 10 * (20/7) = 20%
What the damage boost currently is: 10 * 20 / 7 = 28%

----

Basically, the updater should find the number of increments first and then just apply the percentage that amount of times. Something like:
```
local int increments = heroLevel // stackLevel
val = (int) val * (1 + (increments * percentage))
```
percentage being the decimal representation (so 0.03 instead of 3)

Edit: I have just realized this approach wouldn't work for Adela specifically because her specialty adds a previously nonexistent bonus to the spell. What I do there, is just add a `GENERAL_DAMAGE_PREMY` bonus to the table of bonuses that should be applied by the spell with the magnitude of 10 * increments since it's 10% per n levels, vanilla would use 3.

IvanSavenko:
Ah, yeah, I remember something like that, thank for reminder. Will review specialties code and set up testing to prevent regressions - had it in my plans because I have small suspicion that some spell specialists are broken with lua (unverified)

Are there any other edge cases with specialties in hota, not necessarily related to this case? I remember about Clone special, any other edge cases not present in h3?

Karyoplasma:
The other one-offs are:
+ Cure which increases the effect on stronger creatures more (per 8-n levels where n is creature tier). Added it in this PR.
+ Forgetfulness which increases the number of units that are affected by Forgetfulness per n levels where n is creature tier

I am not 100% sure how Forgetfulness specialty works tho and if it affects how many creatures can shoot too or only how many creatures in a shooter stack cannot use their melee attack on advanced+ (this part of the spell is not in core even tho it's part of SoD and considered a bug there). User commentary on the wiki suggests the latter. Also a simple fix associated with that is that advanced should have `"range" : "X"` in core. It only affecting the target stack on advanced is a HotA change but it's somehow the default behavior in core.

Another thing:
`SPECIAL_SPELL_LEV` does something weird as well, or at least I am suspecting that it is the culprit. It seems as if it somehow adds the bonuses to the spell sequentially with a minimum damage increase of 1 which leads to spell specialists like Deemer doing a bit more damage than they should at high levels against high level creatures. ~(it's like 5 flat more damage from Meteor Shower at lvl 20 against Titans, so this could just be a rounding error)~ Edit: it's a combination of the `TIMES_HERO_LEVEL_DIVIDE_STACK_LEVEL` bug and weird rounding that somehow makes it better. 
 
Found my test results for `SPECIAL_SPELL_LEV` from yore (I never reported it because the difference is miniscule):

lvl 20 deemer 10 Spell Power (275 base damage):
unmodded homm3:
peasants: 440, used formula: 275 * (1 + 0.03 * (20/1))
serpent fly: 325, used formula: 275 * (1 + 0.03 * (20/3))
titans: 292, used formula: 275 * (1 + 0.03 * (20/7))

vcmi (only vcmi-extras enabled) and what presumably happens:
peasants: 440, used formula: 275 * (1 + ((int) (3 * 20 / 1)) / 100), same value because (heroLvl % stackLevel) == 0
serpent fly: 330, used formula: 275 * (1 + ((int) (3 * 20 / 3)) / 100)
titans: 297, used formula: 275 * (1 + ((int) (3 * 20 / 7)) / 100)

Karyoplasma:
I'll leave it open until tomorrow, then merge.

PR #659, opened 2026-07-22
Rewrite spell specialties to the new system

Karyoplasma:
vcmi/vcmi#7621 allows for scaling the specialties without overriding `"core:timed"`. This PR changes the specialists to this new system and removes the Lua override.

Progress:
- [x] Spell casters that scale Damage (Deemer, Solmyr, etc.) work correctly
- [x] Resurrection specialists work correctly 
- [x] Fixed enchant specialists (Mirlanda, Xsi, Aenain, etc.) work correctly
- [x] Percentage scaling enchants (Dargem, Dalton, etc.) work correctly
- [x] Cure specialists work correctly (likely stays patched in Lua)


Karyoplasma:
With this PR, HotA's spell specialty adjustments will not impact other mods. All patches check against custom bonuses (`SPECIALTY_CURE` and `SPECIALTY_CLONE`) and if the hero doesn't have them, behavior is unchanged, so you could make a vanilla-scaling Cure specialist while having HotA activated.

All specialty changes are now implemented. Coronius was the very last holdout with his Slayer specialty, but he is fixed now as well. I have tested the scaling of all spell specialties and they work as expected.

@GeorgeK1ng What we could do, is to get rid of adjusting the global for creature specialists and move that to the config files. These have to exist anyway because we need change the strings there. Then, gamebalance.heroes would not impact other mods at all. I could just do that with this PR, would that be beneficial?

PR #665, opened 2026-07-26
Rewrite creature specialists

Karyoplasma:
This PR aims to get rid of changing the global setting for scaling creature specialists in gamebalance.heroes. This reduces the impact activating the mod has on other mods.

It also adds some campaign heroes that were forgotten to scale with the rest.


Karyoplasma:
@GeorgeK1ng @misiokles 

What do you think about this?

The upside is that other mods now don't have to make a compatibility patch for their creature specialists as it won't affect their scaling anymore. They still can do it to make them scale the same, but it prevents the issue of presenting wrong information to the user: If the text says 5% it's 5%, not actually 20%.

The downside is that existing compatibility patches need to account for that and either change the global themselves or also adjust the scaling of their heroes on a per hero basis. It's not much more work because in order to prevent wrong information display, they would have had to update the strings anyway, so the files would already exist.

misiokles:
Sorry for the late reply.
The mods have been updated many times over the years to keep up with changes in the VCMI engine, so there will be time to keep up with the changes in both the engine and HotA's balance this time as well. So yes - I'm glad you decided to merge this PR.

PR #677, opened 2026-08-10
Water Wheel blocking fix

Karyoplasma:
In mapdecorations, a snow water wheel template is added that uses a different block mask than vanilla (same block mask as the grass version). However this leads to the problem that now the RMG places either the vanilla one with the wrong block mask or the hota one at a 50/50 chance.

This PR stops the entire vanilla Water Wheel object type from spawning and instead introduces a new type "hotaWaterWheel" that contains the templates with the fixed masks. Since this adds an object type, I moved it to mapObjects.

I'm not sure whether this is a good solution. It does have its downsides. It shouldn't break saves or maps at least, but it could lead to unexpected behavior with other mods if a compatibility patch isn't present.

@misiokles @GeorgeK1ng 
Let me know what you think about this.


misiokles:
I'm quite happy about this PR.
I guess, in HotA, WoG or maybe other mods, there are more such issues. There mods were written for older engine version and that time they worked (as close as possible), now they don't. So should be fixed :)
Thanks!

Karyoplasma:
Yeah, it would be easier and cleaner if we could modify the object template from core directly, but for some reason overwriting the .msk doesn't work. Maybe that's just a bug tho.

PR #678, opened 2026-08-11
[Needs testing] Combat script for Runes skill

Karyoplasma:
Requires: vcmi/vcmi#7690

Tested:
- [x] caps are respected
- [x] starting rune levels from Altars work
- [x] animations and bonus icons display correctly
- [x] creatures that attack multiple times only increase rune level once per sequence (active and passive)
- [x] own counterattacks do not increase rune level (because that means we have been hit before)
 
Partially working (hacky way of doing things, but functionally okay):
- [x] battle log messages for achieving rune levels on units
- [x] battle log messages for altar of the runes adding starting rune levels

Known bugs:
- [ ] casts from abilities that allow the unit not to skip its turn should not add rune levels
- [x] playing multiple animations (e.g. after a unit casts) is buggy (plays the rune effect and adds the spell effect to the target, but does not play the spell animation. turned of runes animation until fixed.)
- [ ] getting damaged by spells doesn't increase rune levels
- [x] need to fix weird schema errors that i don't understand 

not sure if those are fixable yet. If someone has an idea, go ahead.


Karyoplasma:
All in all, this is ready and I'm going to merge it after thinking about it a bit more. It's not perfect yet, but it's better than the current runes configuration in every way. All bugs I know of are present in the current config as well, so this PR doesn't add any new ones, but rather fixes some cases like multiattacks increasing rune levels multiple times, counterattacking increasing rune levels, etc.

The only known-to-me bugs remaining are:
- getting damaged by a hero spell should increase rune level by 2
- an attack-counterattack sequence should only add 2 runes levels (currently adds 3 - 1 for attacking, 1 for getting hit)
- rune levels should be granted after a unit's action, not during it
- unit spellcasts that don't skip the unit's turn should not grant a rune level (e.g. Crimson Couatl meditation)

PR #681, opened 2026-08-13
Configuration for newly supported objects

Karyoplasma:
With vcmi/vcmi#7661, Warlock's Lab and Temple of Loyalty are now functional. Also Steel Elves can now only melee attack the targets that block them (they can still choose to shoot a different target).


Karyoplasma:
converted to draft until changes are merged

PR #683, opened 2026-08-18
[WIP] Automaton ability combat script

### LINKED ISSUE #556: Is it possible to implement Automaton's detonation mechanism?
The mechanism works as follows:

Automaton can activate a detonation mechanism, triggered upon death, dealing (90 + 5* quantity) of damage to all adjacent creatures around the user, indiscriminately inflicting damage.

The detonation has a special animation and also grants the "disintegrate" ability.

In the VCMI we can use the "trigger upon death" setting in ON_COMBAT_EVENT, but this bonus configuration may not be complete yet.

For the detonation animation, we can create a new def specifically for the original death animation (when its ability is activated) , allowing it to trigger a spell and use this animation when all units are wiped out. Simultaneously, due to the disintegrate ability, the unit's corpse will also disappear upon destruction. At 200%-300% animation speed, the detonation animation can essentially perfectly cover the death animation.

Several challenges exist:

1. Death detection. Theoretically, any death should trigger the effect: being killed by melee attacks, ranged attacks, or being directly or indirectly killed by spells.

2. The detection of spellcasting upon death. Can it be made to trigger in a ring around the unit? A normal Fireball or Ice Ring is actually a ring around a single tile. We need a ring around two-tile creatures, with two more trigger tiles than a single-tile ring.

3. Currently, creature spellcasting can only be detected using a basic formula like quantity * damage (e.g., Thunderbird) or per unit of strength (e.g., Faerie Dragon). How can we use an initial value + quantity + damage method?

Regarding how to activate it in auto-battle mode? Actually, my previous alternative already allows the AI ​​to use it. Simply put, it involves adding a healing effect (which may heal 1 point of health) to "BattleEffect", but this will definitely cause Automaton to activate when it takes damage.

Karyoplasma:
This PR will add the explosion on death ability of the Automatons and Sentinel Automatons. The only downside, probably, will be that the explosion won't trigger if they die to spells as there is no trigger for that yet. Could easily be added in once it exists.

Fixes #556 

Progress:
- [x] Make Automatons explode on death, dealing 90+5*unitsLastAlive damage to all surrounding units
- [x] Implement Frederick's (and Umender's) special ability of increasing that explosion damage
- [x] battle log messages
- [ ] figure out how to damage multiple units synchronously instead of one after another (doesn't seem possible)
- [x] test possible edge case scenarios


Karyoplasma:
The missing trigger can be added later, like in the Runes script. This is good to merge imo.

PR #700, opened 2026-09-03
Prepare Text Polishing translations for Weblate

rafauke:
## Summary

- rename the Text Polishing source directory from `content/config/englishTextPolishing` to `content/config/english`
- update the five English translation paths in the submod manifest
- bump Text Polishing to 1.1.0

The five source files are unchanged; only their language directory and manifest references move.

## Motivation

Using the conventional `english` language directory allows Weblate components to use these files as monolingual base files without committing a generated `english.json` that repository automation would replace.

Translated language files can be created through the Weblate UI after the components are configured.

## Validation

- parsed the manifest and all five JSON-with-comments source files
- verified every configured English translation path resolves
- byte-compared all moved files against their originals
- tested the repackaged submod locally in VCMI with Polish overrides and English fallback


IvanSavenko:
> consolidate the Text Polishing English strings into the standard content/translation/english.json layout.

This won't work as it is - our automation regenerates english.json files and would delete it in this specific case.  

I think better option would be to rename *directory* to `english`

Creation of empty files is no longer requires - can be done via Weblate web UI.

rafauke:
Ok, done @IvanSavenko

Karyoplasma:
I've looked at the files in the mod again and it was a bit messy. Archive contains files with some changes to the English mod, could you include these in this PR? It only moves some strings around (so heroTexts actually contains all hero texts) and sorts them alphabetically. I never thought this mod would be translated, so I got a bit lazy when I made it lol.

[updatedjson.zip](https://github.com/user-attachments/files/31857508/updatedjson.zip)

rafauke:
Done @Karyoplasma :)

PR #706, opened 2026-09-18
Enable Learning level-up via grantsLevelUp skill tag

ykonovets:
Enables the HotA Learning level-up behaviour once the engine supports it: vcmi/vcmi#7857 adds a generic `grantsLevelUp` secondary skill tag and an `offerCooldown` skill property.

With the tag, each time a hero learns or upgrades Learning (level-up pick, Witch Hut, Scholar, Pandora's Box, University), the hero immediately gains one level, and Learning is not offered for upgrade on that level-up as long as the hero has any other skill to upgrade. This is the rule HotA introduced in 1.7.3 and still has in 1.8.1, the current release (the official changelog lists no Learning change in 1.8.0 or 1.8.1). Verified in-game and by tracing HotA's level-up code on 1.7.3; details in the engine PR.

Only the tag and `"offerCooldown" : 1` are added to the existing `core:learning` override in the Learning submod. Engines without this support ignore both, so the change is safe to merge before the engine side ships.


Karyoplasma:
Thanks. Will merge once upstream merges.

ykonovets:
@Karyoplasma just letting you know the upstream caught up

